PR #75, opened 2014-12-16
Remove a warning

namusyaka:
Just a small fix :)

### DIFF
--- a/lib/rom/boot/base_relation_dsl.rb
+++ b/lib/rom/boot/base_relation_dsl.rb
@@ -8,6 +8,7 @@ module ROM
         @env = env
         @name = name
         @header = []
+        @repository = nil
       end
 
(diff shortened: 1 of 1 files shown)


wikimatze:
Would like to work on the integration, any chance @namusyaka to do some pair programming? With vim and tmux?

solnic:
I'd recommend checking out [lotuskase](https://github.com/solnic/lotuskase) which is based on [dry-component](https://github.com/dryrb/dry-component) to configure rom. It's a PoC and the component stuff is experimental so I'm very open to some feedback. The general goal of dry-component is to simplify managing object dependencies within a system (doesn't matter if it's sinatra, padrino, roda or even rails) by providing interfaces to load needed files in correct order, configuration, booting and resolving *object* dependencies automatically.

Issue #83, opened 2015-01-03
Cannot use methods in block passed to `CommandRegistry#try`


alpardal:
Since the block passed to ```ROM::CommandRegistry#try``` is ```instance_eval```'d, I cannot use methods visible to the caller inside the block. For example:

```ruby
class TasksController < ApplicationController

  def create
    rom.command(:tasks).try { create(task_attributes) }
    redirect_to root_path
  end

private

  def task_attributes
    params.require(:task).permit(:title)
  end
end
```
raises
```ROM::Registry::ElementNotFoundError:  :task_attributes doesn't exist in ROM::CommandRegistry registry```

I think it'd be better if the eval context were a block param:

```ruby
rom.command(:tasks).try {|c| c.create(task_attributes) }
```

What do you think?

splattael:
Hi @alpardal,

actually we could do both depending on the context:

### No args provided -> `instance_eval`

Current behaviour.

```ruby
rom.command(:tasks).try { create(...) }
```

### Args provided -> call with `self`

```ruby
rom.command(:tasks).try { |command| command.create(...) }
```

One possible implementation might be [yield_or_eval](https://github.com/tj/terminal-table/blob/ae5504768a9828615e6194e208b762718dc0fe6c/lib/terminal-table/table.rb#L207..L214).

WDYT?

solnic:
@splattael I love this idea. It's more flexible.

Let me just show you *why* I did it like it is now, here's how I use commands:

``` ruby
class CreateUser
  attr_reader :users

  def initialize(users)
    @users = users
  end

  def call(input)
     users.try { create(input) }
  end
end

create_user = CreateUser.new(rom.command(:users))
create_user.call(params[:user])
```

There a couple of things here:

* I don't use strong params
* I use virtus params object to deal with input (that's set on the create command using `input` dsl)
* I always pass input to the method so it's visible within the block
* I never call anything inside the block except the commands

Having said that - it's just me. I can see how and why people wouldn't prefer to use it differently.

I think yield_or_eval is a really simple and nice change so :+1: from me.

Thoughts? /cc @elskwid @adz

alpardal:
@solnic I agree with your points, it's just that ```instance_eval``` breaks, IMO, the "expected" behavior, and may surprise new users.
@splattael I like ```yield_or_eval```: simple to change and doesn't break the current api.

solnic:
Yep totally, hence :+1: for this change from my side :smile:

splattael:
[...]

PR #84, opened 2015-01-03
Pass command to block in CommandRegistry#try

splattael:
This PR addresses #83

/cc @solnic


elskwid:
Very elegant solution to the problem. Thanks @splattael.

PR #85, opened 2015-01-03
Refactor Options module

splattael:
Move validation code into class.

After this commit including `Options` exposes less class methods:
* `option` and
* `option_definitions`

Feedback is welcome!


solnic:
Thanks. Looks great. In fact I wanted to do the same after the first pass but it was 4am ;) Merging!

PR #91, opened 2015-01-04
WIP adapter interface improvements

solnic:
This branch introduces a couple of improvements that should help in building adapters:

* lint tests for adapters in minitest (which is part of ruby so everybody has it)
* new modules/classes that can be re-used in adapters to unify and simplify implementation of individual adapters


solnic:
Soooo any objections or can I merge this in? /cc @elskwid

lib/rom/array_dataset.rb
```diff
@@ -0,0 +1,12 @@
+require 'rom/enumerable_dataset'
+
+module ROM
+  class ArrayDataset < EnumerableDataset
+    forward(
+      :*, :+, :-, :compact, :compact!, :delete_if, :flatten, :flatten!, :keep_if,
+      :map!, :length, :pop, :reject, :reject!, :reverse, :reverse!, :sample,
+      :select!, :size, :shift, :shuffle, :shuffle!, :slice, :slice!, :sort!,
+      :sort_by!, :uniq, :uniq!, :unshift, :values_at
```
elskwid: @solnic – any feelings on the maintainability of this? (I am *not* suggesting we change it for this PR :grin: )

lib/rom/support/data_proxy.rb
```diff
@@ -0,0 +1,51 @@
...
+    ].freeze
+
+    def self.included(klass)
+      klass.class_eval do
+        extend ClassMethods
+        include Equalizer.new(:data)
+      end
+    end
+
+    def initialize(data, header, tuple_proc = self.class.tuple_proc)
```
elskwid: OH! I like that the proc can be injected *or* set in the class method.

lib/rom/support/data_proxy.rb
```diff
@@ -0,0 +1,51 @@
...
+      @data = data
+      @header = header
+      @tuple_proc = tuple_proc
+    end
+
+    def each
+      return to_enum unless block_given?
+      data.each { |tuple| yield(tuple_proc[tuple]) }
```
elskwid: You know this makes me happy. As I said yesterday, this is what I would expect: the proc applied to the tuple on demand rather than all at once. I suppose then it will play nice with usages like:

```ruby
my_dataset.each.take(5) { |row| do_stuff }
```

So that the proc is only applied the 5 times it's needed.

spec/test/memory_adapter_lint_test.rb
```diff
@@ -0,0 +1,22 @@
+require 'rom/adapter/memory'
+require 'rom/adapter/lint/test'
+
+require 'minitest/autorun'
+
+class MemoryAdapterLintTest < MiniTest::Test
```
elskwid: I saw your changes to the constant. There's something else to keep in mind, Minitest 5 is `Minitest::Test` <- note the 't'. I believe v5.x includes an alias so `MiniTest` will still work.

spec/unit/rom/array_dataset_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+require 'rom/array_dataset'
+
+describe ROM::ArrayDataset do
+  subject(:dataset) { klass.new(data, []) }
+
+  describe 'with array' do
+    let(:klass) do
+      Class.new(ROM::ArrayDataset) do
+        def self.tuple_proc
+          Transproc(:symbolize_keys)
```
elskwid: This use of Trasproc here hints at some cool stuff ...

lib/rom/support/data_proxy.rb
elskwid: I'll just voice my opinion that the name `tuple_proc` doesn't do it for me. I do not have a better name in my head - maybe `row_proc`? Anyway, food for thought as we go forward.

lib/rom/adapter/lint/test.rb
```diff
@@ -0,0 +1,54 @@
+module ROM
+  class Adapter
+    module Lint
+      module TestAdapter
```
elskwid: @solnic the `TestAdapter` is precisely what we discussed. It's great. What a help to developers to be able to wire this up and see if their adapter meets the requirements.

elskwid:
@solnic I only have two items which I believe rise to the level of `objection`. 

1. The explicit forwarding of all the enumerable methods frightens me. How hard will this be to maintain?
2. The name `tuple_proc` isn't great at first glance. Perhaps I'm alone in not immediately thinking about tuples when I see the name dataset. I tend to think of 'rows' and 'columns'. I blame Sequel. In the commit I suggested `row_proc` but if we're talking tuples everywhere else then ... *shrug*

Overall, it's grand. It took shape quite fast from my point of view. Did it help you flesh out how the adapters could be used and built at all? If so, then it was totally worth it. I'd like to add it to `rom-yaml` and `rom-csv` tomorrow morning. (and others?).

lib/rom/support/data_proxy.rb
solnic: Yes, I "blame" Sequel too (I borrowed "dataset" from it). In fact, those may *not be tuples* (like in CSV adapter) so we should come up with a better term here. I guess "row" would be OK since it's "a row in a dataset". This may or may not sound good in a particular adapter's case. For instance "row" works great for CSV obviously.

spec/test/memory_adapter_lint_test.rb
solnic: Ah, yes, CONSTANTS :P

lib/rom/array_dataset.rb
solnic: @elskwid well, we will have to keep an eye on Enumerable and Array methods. we'll see how it goes. Auto-forwarding of all methods would require a more complicated implementation (mostly because of to_enum stuff, otherwise we'd break semantics of enumerable/array). I didn't have fun listing those methods manually trust me, but I think it's the right thing to do. Needs to be explicit otherwise you may get funny stuff going on in your code (for instance we do not want to always wrap response with a dataset object, something we totally want a plain array back)

lib/rom/adapter/lint/test.rb
```diff
@@ -0,0 +1,83 @@
...
+        attr_reader :adapter, :uri
+
+        def test_schemes
+          assert_respond_to adapter, :schemes,
+            "#{adapter}.schemes must be implemented"
+
+          assert_instance_of Array, adapter.schemes,
+            "#{adapter}.schemes must return an array with supported URI schemes"
+
+          assert adapter.schemes.size > 0,
```
splattael: @solnic 

With `minitest` we can use `assert_operator adapter.schemes.size, :>, 0`.
It gives a nice error message IMHO.

Alternatively, we could also use `assert adapter.schemes.any?` to match the assert message better?

solnic: I changed it to `any?`. thanks :)

PR #97, opened 2015-01-12
Add one and one! to Reader

aflatter:
There we go.

### DIFF
--- a/lib/rom/reader.rb
+++ b/lib/rom/reader.rb
@@ -58,6 +58,36 @@ module ROM
       mapper.process(relation) { |tuple| yield(tuple) }
     end
 
+    # Returns a single tuple from the relation if there is one.
+    #
+    # @raise [ROM::TupleCountMismatchError] if the relation contains more than
+    #   one tuple
+    #
+    # @api public
+    def one
+      if relation.count > 1
+        raise(
+          TupleCountMismatchError,
+          'The relation consists of more than one tuple'
+        )
+      else
+        mapper.process(relation).first
+      end
+    end
+
+    # Like [one], but additionally raises an error if the relation is empty.
+    #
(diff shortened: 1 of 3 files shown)


lib/rom/reader.rb
```diff
@@ -58,6 +58,36 @@ def each
...
+    def one
+      if relation.count > 1
+        raise(
+          TupleCountMismatchError,
+          'The relation consists of more than one tuple'
+        )
+      else
+        mapper.process(relation).first
+      end
+    end
```
solnic: That's an interesting situation - this introduces a requirement that every relation must respond to `count` (provided by the adapter). Commands also require this interface but for simple read-only data sources an adapter wouldn't have to care about that. If we use `count` here we'd introduce a new required interface for *all* adapters. Hmmmm

lib/rom/reader.rb
```diff
@@ -58,6 +58,36 @@ def each
...
+    # @raise [ROM::TupleCountMismatchError] if the relation does not contain
+    #   exactly one tuple
+    #
+    # @api public
+    def one!
+      one || raise(
+        TupleCountMismatchError,
+        'The relation does not contain any tuples'
+      )
+    end
```
solnic: IIRC using `or` here is more idiomatic

lib/rom/reader.rb
aflatter: For future reference: We discussed this on Gitter and we'll go with what we have right now, which means that `#one` and `#one!` currently require the relation to implement `#count`. In the future, we may provide a naive implementation using `Enumerator#next` (to check if there's more than one tuple in the relation) and the possibility to provide an implementation in the adapter in order to skip loading all records if it hurts performance.

PR #98, opened 2015-01-12
Remove unused `connection` from Repository

aflatter:
Let's remove this - it's not used, but marked as public API and we're thinking about removing `Adapter#connection`.


solnic:
Yep. Makes sense.

PR #99, opened 2015-01-12
Simplify #repository in BaseRelationDsl

aflatter:
This method changes its behavior after the first call. It might not hurt
right now as `@repository` is always empty, but let's clean this up
anyways.


solnic:
I feel like we need to encapsulate this in some support module

PR #106, opened 2015-01-13
Remove schema DSL

solnic:
This adds ability to specify :repository when defining a relation:

``` ruby
  setup.relation(:users, repository: :postgres)
  # or
  setup.relation(:users) # this uses :default as :repository name
```


elskwid:
@solnic - this is a great bit of surgery. It really highlights how `schema` was no longer needed. From a pragmatic standpoint even if we find out we need schema later it would be better to add it then rather than keep it around when in doubt.

solnic:
Indeed :) I introduce this concept because it made perfect sense when ROM was based on Axiom. Now it was just an overhead and additional complexity (same with header).

solnic:
OK just realized that I don't see anything to add here. Let me know if you see something fishy. Otherwise I'd like to merge it in. I'll follow up with some small refactorings in `Finalize` in master before 0.6.0 release. For now I just want to get that into master.

lib/rom/setup.rb
```diff
@@ -54,8 +33,8 @@ def schema(&block)
     #   end
     #
     # @api public
-    def relation(name, &block)
-      @relations.update(name => block)
+    def relation(name, options = {}, &block)
+      @relations.update(name => [options, block])
```
aflatter: I know it's not the scope of this PR, but I can use a relation name twice and it won't blow up?

lib/rom/setup/finalize.rb
```diff
@@ -8,59 +8,51 @@ module ROM
...
-        end
-
-        base_relations = @schema.each_with_object({}) do |(repo, schema), h|
-          schema.each do |name, dataset|
-            adapter_relation_map[name] = repo.adapter
-            h[name] = dataset
-          end
+      def load_datasets
+        repositories.each_value do |repository|
+          datasets[repositories.key(repository)] = repository.schema
```
aflatter:
```ruby
repositories.each do |key, repository|
  datasets[key] = repository.schema
end
```

lib/rom/setup.rb
solnic: Defining the same relation twice is not supported and it should blow up. I'll improve it in a separate PR

lib/rom/setup/finalize.rb
solnic: DOOOOOH :stuck_out_tongue_winking_eye:

lib/rom/setup.rb
aflatter: Let me open an issue for that so we don't forget.

lib/rom/setup/finalize.rb
aflatter: Refactoring blindness :dart:

lib/rom/setup/finalize.rb
```diff
@@ -8,59 +8,51 @@ module ROM
   class Setup
     # @private
     class Finalize
-      attr_reader :repositories, :adapter_relation_map
+      attr_reader :repositories, :datasets, :adapter_relation_map
 
       # @api private
       def initialize(repositories, schema, relations, mappers, commands)
```
splattael: `schema` is not used anymore.

lib/rom/setup/finalize.rb
```diff
@@ -71,19 +63,18 @@ def load_relations(schema)
       end
 
       # @api private
-      def build_relation(name, builder, block = nil)
-        adapter = adapter_relation_map[name]
+      def build_relation(name, builder, options = {}, block = nil)
+        repo_name = options.fetch(:repository) { :default }
```
elskwid: We should document the `:default` repo functionality somewhere. It has already tripped a few people up when new to rom.

solnic: Definitely

elskwid:
:tada:

PR #112, opened 2015-01-14
Use more Ruby for basic linter

splattael:
## WIP! Needs review!

__Rebased and squashed down. Details are from the commit messages__

Adds a basic linter (`ROM::Lint::Linter`) that wraps up the common needs out of a plain Ruby linter: a list of lints, named like test methods (lint_*), a way to yield to each lint method, and a method to perform a specific lint returning `true` so they are assertion friendly.

Adds specific linters for a repository and a dataset. These can be used by repository adapter authors to verify if they comply with our api.

Updates the Minitest/TestUnit module to use the new `ROM::Lint::Test` modules (`TestRepository` & `TestEnumerableDataset`) to get access to the test method versions of the lint tests.

Use new repository feature linters to create rspec shared examples for use in an rspec suite.

Add shared features to `memory/dataset_spec` and create new `repository_spec` so our linters are running as part of our normal spec suite.


solnic:
Please see my commits in the reverted [commit](https://github.com/rom-rb/rom/commit/f59011fccc06d5ce552891261463ae31851cc28d)

splattael:
Just an idea. How about define lint methods like that:

```ruby
lint "schemes returns any supported scheme" do
  assert "#{adapter}#schemes must return at least one supported URI scheme" do
    adapter.schemes.any?
  end
end
```

Too fancy?

splattael:
I've merged with master. We can squash afterwards, right?

solnic:
Si. I guess this needs some updates after repo/adapter changes.

splattael:
After the Adapter/Repository collapse: Are we still linting adapters; or now repositories? :dancers:

solnic:
I think now "Adapter" describes a library that provides support for a specific datastore. So, yes, this is still linting for adapters, more specifically lint tests for base adapter objects: `Repository` and `Dataset`

splattael:
Merged with fast-moving master.

I'm still not sure if I'm mistaken in 69e383b by ditching `Adapter` namespace.

elskwid:
@splattael as I mentioned in chat, this is going to require a helluva rebase. Are you okay with us losing some detail in the commits? With the Adapter removal it's going to be very difficult to keep it all straight.

splattael:
@elskwid Sure, no problem. Squash it :)

elskwid:
@splattael and @solnic (@aflatter too) - this is ready for a good look-over. :eyes:

phlipper:
This is a really nice little changset, I'd only recommend adding some context to the README.

lib/rom/lint/repository.rb
```diff
@@ -0,0 +1,55 @@
...
+        return if repository.schemes.instance_of? Array
+
+        complain "#{repository}#schemes must return an array with supported schemes"
+      end
+
+      def lint_schemes_returns_any_supported_scheme
+        return if repository.schemes.any?
+
+        complain "#{repository}#schemes must return at least one supported scheme"
+      end
```
solnic: scheme-related checks can go away given that #120 removes it

spec/unit/rom/memory/repository_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'spec_helper'
+require 'rom/lint/spec'
+require 'rom/memory.rb'
```
splattael: `.rb` extension is not needed
`require 'rom/memory'`

elskwid: Good catch!

PR #117, opened 2015-01-15
Replace Adapter with Repository

solnic:
Repository turned out to be an anemic object that does nothing. That's
why adapter class logic becomes what used to be repository. This
simplifies ROM even more by collapsing two concepts into one.

With this design an adapter for ROM simply inherits from abstract
repository class and provides required interfaces.

It also describes the domain better: we have repositories that provide
datasets. Memory adapter becomes an in-memory repository (which sounds
perfect).


lib/rom/adapter/lint/test.rb
```diff
@@ -16,46 +16,46 @@ module Lint
...
 
-      # This is a simple lint-test for an adapter dataset class to ensure the
+      # This is a simple lint-test for an repository dataset class to ensure the
       # basic behavior is correct
       #
       # @example
       #
       #  class MyDatasetLintTest < Minitest::Test
-      #    include ROM::Adapter::Lint::TestEnumerableDataset
+      #    include ROM::repository::Lint::TestEnumerableDataset
```
aflatter: s/repository/Repository

solnic:
Small note: this *removed* around 100LOC related to anemic repository object. Then @splattael added some stuff related to specs so original code-removal stats are not visible. Something to keep in mind :)

lib/rom/repository.rb
```diff
@@ -1,58 +1,188 @@
+require 'addressable/uri'
+
 module ROM
-  # Repository exposes schema if supported by the adapter
+  # Abstract adapter class
```
elskwid: @solnic doc in `ROM::Repository` still refers to `adapter`. Would it be better to refer to repository or make it clear that an adapter is providing this repository?

PR #118, opened 2015-01-15
[WIP] Remove ROM::Config

aflatter:
rom-rails now completely takes care of reading ActiveRecord configuration, so we don't need this in core anymore.


aflatter:
Specs don't pass yet, as `:adapter` is not supported anymore.

aflatter:
@solnic think you asked about this on Gitter.

solnic:
Right :)

solnic:
w00t

PR #120, opened 2015-01-16
Make repository type explicit

aflatter:
An experiment on removing URIs from rom-core and making the adapter type explicit.

## Rationale

It turns out that a lot of database configurations can not be properly encoded in an URI, e.g. providing username and password but no host: `ROM.setup('postgresql://', username: 'user', password: 'password')`.  The most extreme example is if you rely on the defaults of Sequel completely and you'll have to pass `postgres://` as an URI. Another issue is that if you want to pass an object into the repository, you'll always have to use the options hash anyways.

The change will make repository configuration slightly longer:

```ruby
ROM.setup('mysql://user:pass@@host/database')
# vs.
ROM.setup(:sql, 'mysql://user:pass@@host/database')
```

Blessed (official) adapters will be accessible via a Symbol as shown above. Additionally, it will become possible to pass in repository instances, making the use of custom repositories easier.

```ruby
ROM.setup(MyRepository.new(...))
```

## TODO

- [x] Remove `uri` and `initialize` from `ROM::Repository`
- [x] Get rid of scheme registry, use a simple map instead
- [x] A user-friendly error when using the old style
- [x] Move rom umbrella repositories map to a constant
- [x] Replace `UMBRELLA_REPOSITORIES` with a registry
- [x] raise `LoadError` with a nice message when a repository type can't be loaded
- [x] Rubocop clean?


spec/integration/mappers/deep_embedded_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'rom/adapter/memory'
 
 describe 'Mappers / deeply embedded tuples' do
-  let(:setup) { ROM.setup('memory://test') }
+  let(:setup) { ROM.setup(:memory, 'memory://test') }
```
aflatter: @solnic all of these will turn into `ROM.setup(:memory)` which is probably a sign that the change is a good thing.

solnic: Lovely!

solnic:
Scheme registry won't be needed. We should consider having either a convention (`:sql` => `ROM::SQL` `:memory` => `ROM::Memory` etc.) and automatically infer constant from symbol or make it explicit using a map and registration mechanism (like with schema but using constant symbol identifiers)

aflatter:
Right, but we'd need to also include the path to the repository in the map, plus we can't rely on the constant being defined. `{sql: ['ROM::SQL', 'rom/sql']}`

solnic:
@aflatter we can auto-require them (which is funny as it would go back to how it used to be in ROM hah)

solnic:
`:sql` == `require 'rom-sql'` and `ROM::SQL` is expected to be defined afterwards

aflatter:
That is simple and would work with our repositories.

solnic:
:+1:

aflatter:
@solnic @elskwid @splattael so after a heavy rebase, I think this is ready to merge.

splattael:
Nicely done! :+1: 

As long as rubocop passes :shipit:

solnic:
Rubocop doesn't pass on master currently :)

lib/rom/repository.rb
```diff
@@ -19,67 +19,35 @@ class Repository
...
-      registered.detect { |repository| repository.schemes.include?(scheme.to_sym) }
-    end
-
-    # @api private
-    def initialize(uri, options = {})
-      @uri = uri
-      @options = options
-      setup
+      klass_name = UMBRELLA_REPOSITORIES.fetch(type)
+      ROM.const_get(klass_name).const_get('Repository')
```
solnic: I think we should just do `ROM.const_get(Inflecto.classify(type).const_get('Repository')` here and remove `UMBRELLA_REPOSITORIES` there's no need to have them listed here.

spec/integration/mappers/embedded_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'rom/memory'
 
 describe 'Mappers / embedded' do
-  let(:setup) { ROM.setup('memory://test') }
+  let(:setup) { ROM.setup(:memory) }
```
solnic: This is awesome :)

lib/rom/repository.rb
```diff
@@ -19,67 +19,35 @@ class Repository
...
-      uri = Addressable::URI.parse(uri_string)
-      klass = self[uri.scheme]
-
-      unless klass
-        raise ArgumentError, "#{uri_string.inspect} uri is not supported"
+    def self.class_from_symbol(type)
+      begin
+        require "rom/#{type}"
+      rescue LoadError
+        fail ArgumentError, "#{type.inspect} is not supported"
```
solnic: I guess it'd be nicer to have an explanation like `Failed to load rom-foo adapter`. This can happen when you simply forgot to add rom-foo to Gemfile etc.

lib/rom/repository.rb
elskwid: Is the hash helping us translate from `csv` -> `CSV` and `memory` -> `Memory` since casing may differ?

spec/integration/repositories/extending_relations_spec.rb
```diff
@@ -2,12 +2,8 @@
 
 describe "Repository" do
   include_context "users and tasks" do
-    before(:all) do
+    let(:repository_class) do
```
elskwid: Ah, nice. I went after this `before(:all)` in #125 as well.

spec/integration/mappers/embedded_spec.rb
elskwid: :clap: :clap:

elskwid:
@aflatter I really like what you've done here. Nice work.

solnic:
I've given some second thoughts re adapter registration (esp after seeing #125 and issues we had in specs). Then I had third thoughts, and fourth too.

Eventually I realized that auto-registration:

- introduces implicit behavior that may have "unexpected" side-effects causing "wtf" moments for library users and adapter creators
- makes it hard to find constants based on symbol identifiers
- makes our spec suite more complicated

All this in order to avoid having to write `Repostiory.register(:sql, SQL)` - so, I'd vote for bringing back explicit registration like it was before to simplify things.

solnic:
Ugh, I confused two things: auto-registration (that is gone already here) and linking a symbol identifier to a constant - so, my point is: this should be done via explicit registration maybe something like `ROM.register_adapter(:amazing_db, ROM::AmazingDbAdapter)`

aflatter:
@solnic are you talking about repositories provided as part of the ROM project or 3rd-party/custom repositories or both?

lib/rom/repository.rb
aflatter: True! :+1:

solnic:
@aflatter I'm talking about registering an adapter *namespace* under a specific symbol so that we don't have to do any inference or pollute rom with symbol=>adapter maps

lib/rom/repository.rb
aflatter: @elskwid Yes, that's the reason. That's also why I don't think @solnic proposal works, though I'm not familar with Inflecto.

solnic: Exactly the reason why I proposed explicit registration in the comment below

aflatter:
@solnic Yes, but would that be part of the public API or not? If yes, why?

solnic:
@aflatter yes, public, for adapter creators, to make things explicit, reduce complexity in rom and avoid wtfs.

I also realized that it wouldn't help *much* with spec suite situation - we still need a way to unregister adapters when we dynamically create them for testing purposes, but I'd just have a private api for that in spec/support and call it a day

solnic:
Just a heads-up that I'll wrap this up later today. Stay tuned.

PR #121, opened 2015-01-16
Move ROM::Adapter::Memory to ROM::Memory

aflatter:
Aligns the memory repository with how all others are organized.


solnic:
I like it. My only concern is that "Memory" is such a generic word that it might be hard to figure out that it's actually an in-memory adapter. Hmmm

aflatter:
I agree. Also, Lint is still in `ROM::Adapter` which is now the only thing using the word `adapter`.

solnic:
I wouldn't want to use `Adapter` (or `Adapters`) namespace for adapters, specific db support for ROM can include many things, I'm not even sure "adapter" is a good word here hah.

Issue #122, opened 2015-01-16
Make it possible to cherry-pick a specific mapper


solnic:
Basically making this possible:

``` ruby
 # this means there's no need to define ":de" relation within base translations one
rom.read(:translations, mapper: :de).to_a
```

solnic:
@aflatter after our recent discussion I believe I should revert it to the original idea of `rom.read(:base_rel_name, mapper: :some_mapper)`. OK?

aflatter:
@solnic ack.

solnic:
@aflatter one more thing - we talked about an alternative interface where you can use a block where you put together a relation and then tell which mapper should load it, IIRC something like:

``` ruby
rom.map(:users, :mapper_name) { active.admins.by_id(1) }.one!
```

Right?

aflatter:
I think our ideas were more like this:

```ruby
# Option 1
rom.read(:mapper_name) { users.active.admins.by_id(1) }.one!
# Option 2
rom.read { users.active.admins.by_id(1) }.map(:mapper_name).one!
```

Note that `users` and the rest of the relation are close together, whereas in your example it's somewhat disconnected. Not sure why the method is called `map` in your example, but I guess it has something to do with your concern about encapsulating data access.

solnic:
Ah yeah that's better. So I guess for `map` to work we'd do something like `if block_given? then super else mappers[mapper_name].process(relation).to_a } end`. That block would be evaluated on some object that has access to relations (we could build it during setup to avoid method_missing). Hmmm

solnic:
@aflatter I'd like to remove instance_exec and yield a relation to the block and then close this issue, so:

``` ruby
rom.read(:users) { |users| users.by_name('Jade') }.map(:foobar)
```

lemme know if that works for you, if not let's talk this through :) I'd like to close this chapter for 0.6.0

aflatter:
It's a bit longer, but easy to understand. :+1:

solnic:
@aflatter we could do the trick with arity :)

PR #123, opened 2015-01-16
Add ability to provide a mapper name in Env#read

solnic:
refs #122
cc @aflatter


solnic:
that would couple relation with mapper :( suddenly it would know about its existence

solnic:
BTW I should mention a couple of things just to clarify:

- `rom.relations` are exposed because of a lack of certain interfaces; eventually I would like to have some higher-level interfaces exposed by `env` so that you don't have to access relations directly. For instance I can easily see a separate interface for aggregations (max, min, count etc.)
- `Reader` object is a top-level interface for reading defined relations via mappers. It is an enumerable object that only exposes publicly available relations and *excludes native query DSLs or any other interfaces provided by the adapter layer*

I have to admit this my little experiment and maybe this approach will turn out to be too strict and not *flexible* enough (being enforced to specify relations for every part of your system that needs data can be annoying in very simple applications). So this is something that might be changed. Regardless - I like reader idea as a higher-level object used only for accessing relations, hence I'd like to have this feature to be part of the reader API, not relation. Especially that, like I already mentioned, Relation doesn't know about the mappers and I believe it should stay like that.

solnic:
OK so now it's:

``` ruby
rom.read(:users).with_mapper(:prefixer)
```

HAPPY?!!??!?!?! ;)

aflatter:
Haha, let's go with that :+1: Thanks for explaining, you successfully managed to convince me not to :fire: reader. ;-)

solnic:
:relieved:

PR #125, opened 2015-01-19
Repository adapter registration

elskwid:
## Problem
In my work with the lint tests I've struggled with the global state created by the adapter auto-registration with the `inherited` hook. It can cause unexpected behavior like adapters registered from a shared context and the need to control the requiring of `rom/memory` to prevent or control the order of registration.

## Proposal
I have brought back (I think there was a way to register adapters before...) an explicit `register` / `unregister` as well as a class level flag to enable or disable the auto-registration. Provides a way to turn off the auto-registration of repository adapters when they inherit from ROM::Repository.

Using this will mean our specs may be more explicit but it also means we can control the environment and test different behavior in isolation without workarounds and complications.

## Impact
* Four methods on `Repository`
* An additional class variable 
* More explicit registration in our specs

### DIFF
--- a/lib/rom/repository.rb
+++ b/lib/rom/repository.rb
@@ -26,9 +26,39 @@ module ROM
     # @api public
     attr_reader :options
 
+    # Register a repository adapter
+    #
+    # @example
+    #
+    #   Repository.register(MyRepositoryAdapter)
+    #
+    # @param [Repository] klass
+    #
+    # @return [Array] registered adapters
+    #
+    # @api public
+    def self.register(klass)
+      Repository.registered.unshift(klass)
+    end
+
+    # Unregister a repository adapter
+    #
+    # @example
+    #
+    #   Repository.unregister(MyRepositoryAdapter)
+    #
+    # @param [Repository] klass
+    #
+    # @return [Repository]
+    #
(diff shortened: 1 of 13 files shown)


elskwid:
@solnic, @aflatter, @splattael - take a look and tell me your thoughts. The names and implementation are totally up for discussion. I really wanted to capture my idea while it was causing me pain.

splattael:
@elskwid Interesting. How does your PR conflicts with #120. Do we still need scheme registration after #120 has been merged?

elskwid:
Good catch @splattael. I totally missed that the auto-registration was gone in that PR. It would seem my pain will go away with #120.

solnic:
I [commented](https://github.com/rom-rb/rom/pull/120#issuecomment-70466397) on #120 which is related to this PR too

solnic:
We talked about this with @aflatter today and our consensus is that *our official adapters* would register themselves through a private API and the rest of the adapter world would require passing in an object to `.setup` as we don't want to make any assumptions about the object construction.

Thoughts?

solnic:
I'm going to close this one since after #120 is merged this becomes obsolete

PR #127, opened 2015-01-19
No need to send Module#include as it is public

splattael:
Dear Travis,

please confirm my assumption that `Method#include` is public.

Love,
  Peter


splattael:
It is public. Since Ruby 2.1. Bad luck.

aflatter:
:partly_sunny:

solnic:
Haha unlucky

Issue #131, opened 2015-01-21
Clean up setup process and simplify DSL implementation


solnic:
There's a potential to clean up how ROM sets up its environment and all the objects (relations, mappers and commands). This requires to ensure that:

* ROM exposes abstract classes (or modules) that can be used in your classes to easily configure them using simple class-level DSLs
* It is *easy* to instantiate objects using classes extended with ROM interfaces
* Setup DSL can simply use classes that inherit core DSLs to instantiate objects and store them in the registries

Here's some code to explain this better:

``` ruby
setup = ROM.setup(:memory)

# simple relation class definition extended with rom-sql DSL
class UserRelation < ROM::Relation
  include ROM::SQL::Relation

  one_to_many :posts

  def sorted_index
    order(:id, :name).select(:id, :name, :email)
  end
end

# easy to instantiate a relation object
user_relation = UserRelation.new(setup.repositories[:default].dataset(:users))

# simple mapper class definition
class UserMapper < ROM::Mapper
  attribute :name, from: :user_name

  wrap address: [:street, :zipcode, :city]
end

# easy to instantiate a mapper by passing mapping processor type
user_mapper = UserMapper.build(:transproc) # .new would receive a built header with transformer object

# simple Reader class definition
class UserReader < ROM::Reader
  expose :sorted_index 
end

# easy to instantiate a reader
user_reader = Reader.new('users', user_relation, users: user_mapper)
```

Things to consider:

* Should we drop setup DSL completely and use a typical approach with class inheritance and use inherited hook to register all classes so that finalize process knows what to instantiate?
* Would it be possible to simply use modules from adapters to extend relation classes (ie `include ROM::SQL::Relation` - right now setup uses repository *object* to inject custom interfaces but that might not be needed
* Have a strict setup process where we build objects in specific order without delaying it (like it is done now) - this would get rid of the need to have finalization process. It would make it easier to see when things go wrong as it happens immediately close to your code. Probably the only challenge is to figure out how to expose relations to other relations (reminder: right now every relation has access to all other relations).

solnic:
I'm gonna take a stab at this and see how it "feels"...stay tuned

PR #133, opened 2015-01-25
Improved setup with subclassing

solnic:
### Description

This PR adds an ability to simply define subclasses of core ROM components - `Relation`, `Mapper`, `Reader` and `Command::*` - once defined the setup process will instantiate them and store in the registries just like when using setup DSL.

The new system will be used internally by the setup DSL too which will simplify implementation.

Setup DSL is a subject of an ongoing discussion - the way I see it now is that it's pretty much "a sinatra-style ROM" which actually has its benefits - for a one-file-setup to do something quickly having a DSL can be really useful.

### TODO

- [x] port relation-builder logic to `Relation` class
- [x] port mapper-builder logic to `Mapper` class
- [x] port reader-builder logic to `Reader` class
- [x] port command DSL logic to `Command` classes and update `Setup` and `Finalize` to use it
- [x] refactor `ROM::Setup` and `ROM::Finalize` to use subclassing
- [x] figure out how to extend relation sub-classes with adapter-specific interface

This addresses #131


solnic:
@aflatter @elskwid @splattael this is ready...mind taking a look at the insane DIFF?

spec/unit/rom/reader_spec.rb
```diff
@@ -10,6 +10,55 @@
...
+
+      expect(relation).to receive(:all)
+        .with(1, &block)
+        .and_return([joe])
+
+      expect(mapper).to receive(:process)
+        .with([joe])
+        .and_yield(joe)
+
+      result = reader.all(1, &block)
```
kchien: There appears to be a relationship between the the integer `1` here and on line `38`, but would it be clearer to extract it to an explaining variable within this example?

lib/rom/command.rb
```diff
@@ -0,0 +1,82 @@
...
+      case type
+      when :create then commands.const_get(:Create)
+      when :update then commands.const_get(:Update)
+      when :delete then commands.const_get(:Delete)
+      else
+        raise(
+          ArgumentError,
+          "#{type.inspect} is not a supported command type for #{inspect}"
+        )
+      end
```
kchien: Could this be refactored into a `Hash` lookup, so that the duplication could be removed? The keys would be the symbols (`:create, :update, :delete`) and perhaps the values could be a be a proc representing the code in the `then` clause. If the key isn't found, the error can still be raised.

solnic: I agree this needs some clean up. I'll add a TODO to this code and we could polish in master after it's merged in :)

spec/unit/rom/reader_spec.rb
solnic: Ah, gotcha. I'll use a local var to make it read better.

solnic:
OK sorry to say this but there's one more thing todo here...see PR's description. Damn...

solnic:
FYI I updated my app without any problems. It works both with DSL-based setup and sub-classing. That's reassuring :)

solnic:
OK that's it folks...please please take a look. I'd love to merge this in later today if you can give me :+1: 

It was tested with my app too and rom-sql has been already updated (see rom-rb/rom-sql#9)

elskwid:
Nice work @solnic! :clap: :clap:

Issue #137, opened 2015-01-31
Add ability to register a relation under a different name than base_name


solnic:
Right now `base_name` is used as the name under which a relation will be registered in the env, this makes it not possible to register multiple relations pointing to the same `base_name` (in sql words, a database table). We could easily make it possible to register multiple relations pointing to the same `base_name` by using a new interface ie `register_as` (which is already used for commands btw).

So, it could look like that:

``` ruby
module Relations
  class Users < ROM::Relation[:sql]
    base_name :users
  end
end

module Admin
  module Relations
    class Users < ROM::Relation[:sql]
      base_name :users
      register_as :admin_users
    end
  end
end
```

aflatter:
@solnic probably a good time to rename base_name to dataset, wdyt?

solnic:
@aflatter dammit I forgot about this :D yeah `dataset` is much better and describes what it actually is...we should rename it

PR #140, opened 2015-02-01
Can register relations under a custom name, defaults to base_name

mcls:
Took a stab at #137 

I'm not very familiar with your test suite, so if this needs to be tested in other places, please tell me :smile:

### DIFF
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -24,7 +24,7 @@ module ROM
     include Charlatan.new(:dataset)
     include Equalizer.new(:dataset)
 
-    defines :repository, :base_name
+    defines :repository, :base_name, :register_as
 
     repository :default
 
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -52,7 +52,7 @@ module ROM
       @__registry__ = registry
     end
 
-    def self.register_as(value = ROM::ClassMacros::Undefined)
+    def self.register_as(value = Undefined)
       super || base_name
     end
 
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -52,6 +52,10 @@ module ROM
       @__registry__ = registry
     end
 
+    def self.register_as(value = ROM::ClassMacros::Undefined)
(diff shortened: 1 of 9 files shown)


spec/integration/repositories/extending_relations_spec.rb
```diff
@@ -47,4 +47,19 @@ def freaking_cool?
...
+          register_as :users
+        }
+        Class.new(ROM::Relation[:memory]) {
+          base_name :guests
+          register_as :admins
+        }
+      end
+    end
+  end
 end
```
solnic: This should live in `spec/unit/rom/setup_spec.rb`

mcls: Ok, I'll update that

lib/rom/relation.rb
```diff
@@ -42,6 +42,10 @@ def initialize(dataset, registry = {})
       @__registry__ = registry
     end
 
+    def self.register_as(value = ROM::ClassMacros::Undefined)
+      super || self.base_name
+    end
+
```
solnic: Yep this is what we want. This reminds me - could you also move `Undefined` to `ROM::Undefined`? We should use it in many places :)

mcls: :+1:

mcls:
I just noticed that the `RelationAlreadyDefinedError` only gets raised when using the setup DSL.
Should I move that to `ROM::Setup::Finalize`?

https://github.com/rom-rb/rom/blob/85226fae8624e278703175aa1c23ceabff2e9c74/lib/rom/setup_dsl/setup.rb#L21

solnic:
Yep good idea to move it :+1:

lib/rom/support/undefined.rb
```diff
@@ -0,0 +1,3 @@
+module ROM
+  Undefined = Object.new.freeze
+end
```
solnic: ah, one small thing - could you move that to `rom.rb`? doesn't really deserve to be in a separate file (and requires are slow :))

Issue #142, opened 2015-02-02
Remove descendants tracking


aflatter:
Opening an issue here to discuss the effects of descendants tracking and the removal thereof. Gerenally, I'd say that descendants tracking is undesirable because it's a class-level interface and hidden from the user. Currently it's used to populate the registries of relations, mappers and commands automagically.

Here's a couple of points to discuss:

* Is it currently possible to require ROM objects before calling `ROM.setup`?
* Will issues arise when using multiple environments?
* What's the consequences for the way readers are currently working? AFAIK, a chain of calls to the Reader simply builds a path of said calls, e.g. `users.with_posts.active`, that is then used to retrieve the appropriate mapper.

The alternative to using descendants tracking is to make registry explicit. Note that this won't bother users of `rom-rails` or similar integration gems as they would ship with support and conventions for location and name of those classes.

```ruby
require 'my_app/relations/users'
setup = ROM.setup(...)
setup.register_relation(:users, MyApp::Relations::Users)
```

The advantages are the removal of a fragile technique, independence of load order and setup time, less hassle in the test suite and simplification of ROM core.

/cc @elskwid @solnic @splattael: We already talked quite a bit about this. :)

solnic:
Yeah I'm :+1+ here esp after adding silly code that clears the descendants arrays in specs and rom-rails to_prepare. As you point out this can be done automatically on an integration level with frameworks so it's fine. It would be worth to consider pulling that code into rom core though so that we don't have to do it all over again in every framework integration.

solnic:
@aflatter are you interested in making that change? I have other stuff on my plate that I *need* to do so this is a nice-to-have for me ;)

aflatter:
Sure, I am, but can't give an ETA. :-(

solnic:
Move to 1.0.0 milestone then?

solnic:
No pressure whatsoever but do you think you could wrap this up this week so that we can include it in 0.6.0?

aflatter:
I'll try, but I'm behind schedule as I got struck down by a flue. :-(

solnic:
Sorry to hear that. I'm struggling with an infection myself :/ Seems like we're gonna postpone 0.6.0 release a little bit.

schmurfy:
That's funny because while looking at the example in the README I was wondering how was everything linked together ;)

[...]

PR #145, opened 2015-02-07
[WIP] Remove descendants tracking

aflatter:
WIP of #142

### DIFF
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -57,36 +57,6 @@ module ROM
       Inflecto.underscore(name).gsub('/', '_').to_sym
     end
 
-    def self.registry(repositories)
-      registry = {}
-      descendants = self.descendants.select { |klass| klass.superclass != self }
-
-      descendants.each do |klass|
-        # TODO: raise a meaningful error here and add spec covering the case
-        #       where klass' repository points to non-existant repo
-        repository = repositories.fetch(klass.repository)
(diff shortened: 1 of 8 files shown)


solnic:
For now I did [this](https://github.com/rom-rb/rom/commit/68789c76870068432140a3c20678296e14aa678b) - lemme know what you think. I no longer feel super strong about removing inherited-hook "magic" anymore since descendants are part of the setup now.

solnic:
@aflatter I'd like to close this one, :+1: or :-1: ?

aflatter:
I'd say we'll leave it open for now.

solnic:
I believe we gotta think about something more generic here. I see the need for a feature where ROM during setup exposes some interface for gathering information needed to finalize defined classes. I believe this would become even more important when we have some plugin interface in place too.

Maybe we should gather information in a separate issue to see what is needed. The first one is obvious: a list of defined classes.

aflatter:
We talked a little bit about this on Gitter and I think the preferred direction is to have integration gems like `rom-rails` to derive the defined classes from conventions (or policies in yaks speak), right?

solnic:
Yes, I love the policy idea. Would love to have something like that in ROM. Various objects could use different policies, policies could be configured using some top-level interface exposed by `ROM`

solnic:
This is...obsolete. the main idea with explicit registration is still there, I'm gonna describe it in a new issue. Closing this one.

PR #146, opened 2015-02-08
bubble up nested values

gotar:
Do not allow ROM::Commands::Result#value to nest another instance of ROM::Commands::Result. Bubble up value to root.


solnic:
Merged :)

Issue #104, opened 2015-01-13
Extend Command API with error/success handling


solnic:
Idea for an api:

``` ruby
result = rom.command(:users).try { create(name: 'Jane'  }

result
  .on_error { |err|
    # handle err
  }
  .on_success { |user|
    # do your stuff
  }
```

aflatter:
I'd prefer the naming to not be callback-oriented. You try to run a command and you get back a result which is simply a value.

aflatter:
```ruby
result = rom.command(:users).try { create(name: 'Jane') }

result.or_else { |error| ... }
      .and_then { |result| ... }

result.value # fails if command was not succesful
```

aflatter:
Maybe Rust's `std::result` can serve as inspiration here: http://doc.rust-lang.org/0.11.0/std/result/type.Result.html

solnic:
Yeah I like that. It's closer to `try` that we already have.

elskwid:
I like the Rust `std::result` more than I like `on_error`, `or_else`, `and_then`. Agreed that `on_error` feels more like standard callbacks, and `or_`/ `and_` seems to denote an order of operation as opposed to simple error handling.

In the interest of listing options: Perhaps some of the Promise implementations can serve as inspiration:

https://github.com/petkaantonov/bluebird
https://github.com/tildeio/rsvp.js

```ruby
result = rom.command(:users).try { create(name: 'Jane'  }

result
  .then { |user|
    # user stuff
  }
  .then { |user|
    # other user stuff
  }
  .catch { |err|
    # handle error
  }
```

solnic:
Damn, this is nice too. Hmmmm

snusnu:
what is the motivation behind adding additional api to the result object? imo, as long as the result object encapsulates associated context (i.e. the value in case of success or an error object/symbol in case of failure), rom users are free to handle/integrate their error handling how they like

solnic:
To establish conventional way of dealing with result objects. Notice that it wouldn't enforce the usage. You can still handle result as you like. This applies to commands themselves. You can access them directly and call them and catch errors yourself and so on.

snusnu:
fair enough, i just wanted to question *adding* api that's not strictly necessary. fwiw, i think that if such api is added, it should help with dispatch, i.e. something like the suggested `#then` which yields the result value "unconditionally" wouldn't really help, would it? If you dispatch on success vs. error state tho, you save yourself a conditional in client code.

[...]

PR #149, opened 2015-02-10
Add Result#and_then and Result#or_else as per #104

pdswan:
Considered adding runtime checks to make sure the value returned from the block is a `Result` as per the type signatures in the [Rust docs](http://doc.rust-lang.org/0.11.0/std/result/index.html), but decided to poll opinions first.

### DIFF
--- a/lib/rom/commands/result.rb
+++ b/lib/rom/commands/result.rb
@@ -4,14 +4,12 @@ module ROM
     #
     # @public
     class Result
-      # Wrap a value in a Result type if it's not already wrapped
-      #
-      # @return [Success, Failure]
-      #
-      # @api public
-      def self.wrap(value)
-        return value if value.kind_of?(Result)
-        new(value)
+      def self.success(value)
+        Success.new(value)
+      end
+
+      def self.failure(error)
+        Failure.new(error)
       end
 
       # Return command execution result
@@ -34,11 +32,6 @@ module ROM
       end
       alias_method :to_a, :to_ary
 
-      def wrap(*args)
-        self.class.wrap(*args)
-      end
-      private :wrap
-
       # Success result has a value and no error
       #
       # @public
@@ -52,14 +45,14 @@ module ROM
         #
         # @api public
         def >(other)
-          other.call(value)
+          other.call(value, Result)
         end
 
         # Yield to block with value of result
         #
         # @api public
         def and_then(&blk)
-          wrap(self > blk)
+          self > blk
         end
 
         # Ignore block and return self
@@ -112,7 +105,7 @@ module ROM
         #
         # @api public
         def or_else(&blk)
-          wrap(blk.call(error))
+          blk.call(error, Result)
         end
 
         # Return the error
--- a/lib/rom/commands/result.rb
+++ b/lib/rom/commands/result.rb
@@ -40,6 +40,22 @@ module ROM
           other.call(value)
         end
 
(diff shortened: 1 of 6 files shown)


solnic:
Oh lovely! Could you also write an integration spec showing off how it looks? Maybe in `spec/integration/commands/try_spec.rb`?

pdswan:
Sure thing! Will do that today.

Can use examples to highlight what happens if your block doesn't return a `Result` as well =)

pdswan:
Writing the integration specs made it very clear that not doing *something* about the return values of the blocks passed to `and_then` and `or_else` makes for a pretty poor API.

For now I combined the functionality of Rusts `and_then` with `map` and `or_else` with `map_error`; i.e. if the block returns something that is not of type `ROM::Command::Result` it is automatically wrapped in one.

I'm not a huge fan of the magic of this and I'm sure there are some issues with implementing monadic bind and map as the same function, but I can't think of any right now and I like this better than having to construct a `ROM::Command::Result` every time you use want to use `#and_then`.

aflatter:
Just some ideas.. The challenge is to come up with an API that doesn't feel too alien in the Ruby world.

```ruby
# Option 1
result.and_then do |value|
  if value % 2 == 0
    Error("An even number")
  else
    Ok(value)
  end
end

# Option 2
result.and_then do |value, err, ok|
  if value % 2 == 0
    err["An even number"]
  else
    ok[value]
  end
end

# Option 3
result.and_then do |value|
  if value % 2 == 0
    [:err, "An even number"]
  else
    [:ok, value]
  end
end
```

solnic:
I think option 2 looks good, it's explicit and doesn't require defining global functions. Auto-wrapping does feel like magic that's happening behind the scenes and when you look at `and_then/or_else` blocks it's not obvious that it's going to happen.

pdswan:
Ok, I implemented something close to number 2 but hung both `#success` and `#failure` off of a single passed in dependency:

```ruby
result.and_then do |value, r|
  if value % 2 == 0
    r.failure "An even number"
  else
    r.success value
  end
end
```

Currently, `r` is just `ROM::Commands::Result` to which I've added `.success` and `.failure`.

For kicks, I added a spec for how this affects the implementation's adherence to the monad laws.

It's an easy step away from 2, but now I'm tempted to go all the way and just require folks to do:

```ruby
result.and_then do |value|
  if value % 2 == 0
    ROM::Commands::Result.failure "An even number"
  else
    ROM::Commands::Result.success value
  end
end
```

If we couple this with a runtime error that checks the return value of the block passed to `and_then` or `or_else` we should end up with something pretty user friendly.

But...as I look at this more I think that with [Rust as the example](http://doc.rust-lang.org/0.11.0/std/result/type.Result.html#method.map) `map` and `map_error` may be better for the use case described in #104, especially since I'm a little fuzzy as to whether a `ROM::Commands::Result::Failure` is intended to be manually instantiated without a corresponding `ROM::CommandError`.

I fear I may be in over my head at this point, but I'm happy to continue plugging away/experimenting if you all think it's moving in a valuable direction @solnic @aflatter.

Thanks for entertaining the pull request, very excited about this project!

solnic:
I like yielding result wrapper object to the block; definitely don't like exposing class methods (I'm anti-class interfaces in general).

`Result` could be promoted to be more generic, not strictly related to commands, so that command API can use it, and users can use it too when handling their own error/success cases.

PR #150, opened 2015-02-11
Add support for setting decorators for a reader

solnic:
I found that to be super nice so I thought it could be part of ROM. This basically allows you to decorate mapped objects with additional behavior that can be anything, ie:

* UI view presenters (for presenting *values*)
* UI view renderers (for rendering html chunks etc.)
* JSON serializers (for more higher level stuff like hypermedia etc.)
* probably more...

I don't have a strong opinion if it should be part of rom core or an external plugin. My tactic right now is to put stuff into ROM and extract later once we see things clearly.

### DIFF
--- a/lib/rom/reader.rb
+++ b/lib/rom/reader.rb
@@ -5,8 +5,13 @@ module ROM
   #
   # @api public
   class Reader
+    extend ClassMacros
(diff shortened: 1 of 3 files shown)


solnic:
Closing in favor of #155

PR #155, opened 2015-02-14
Add Relation::Loaded and use it in rom.relation {}

solnic:
This does a couple of things:

- adds `ROM::Env#relation` that yields *relation object* to the block and returns an instance of `Relation::Loaded` it means that you have access to all methods exposed by a relation there including adapter-specific query DSL
- `Relation::Loaded` exposes `#map_with` interface that accepts name of a mapper that should be used
- `Relation::Loaded#map_with` supports multiple mapper names so we can map and decorate tuples passing them through a "mapping pipe", see my comment with an example in this PR
- the contract is that `map_with` can be called only once, it returns an array
- `ROM::Env#mappers` returns a registry of relation mappers that were defined during setup

Things to consider:

- Deprecate usage of Reader and remove it in 0.7.0 (or 1.0.0!) - I realized this is a very heavy abstraction and even though I still want to maintain the logic of restricting access to relation only to methods defined by me and avoid using query dsls I no longer think rom core is the right place to put it
- Extend Mapper API with composability so that we can "join" mappers as a nice lower-level API and re-use it in `Relation::Loaded`

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ end
 # could define it explicitly
 class UserMapper < ROM::Mapper
   relation :users
+  register_as :entities
 
   model User
 
@@ -89,7 +90,7 @@ rom.command(:users).create.call(name: "Joe", age: 17)
 rom.command(:users).create.call(name: "Joe", age: 17)
 
 # reading relations using defined mappers
-puts rom.read(:users).by_name("Jane").adults.to_a.inspect
+puts rom.relation(:users) { |r| r.by_name("Jane").adults }.as(:entities)
 # => [#<User:0x007fdba161cc48 @id=2, @name="Jane", @age=18>]
 ```
 
--- a/lib/rom/env.rb
+++ b/lib/rom/env.rb
@@ -1,9 +1,11 @@
+require 'rom/relation/loaded'
+
 module ROM
   # Exposes defined repositories, relations and mappers
   #
   # @api public
   class Env
-    include Equalizer.new(:repositories, :relations, :readers, :commands)
+    include Equalizer.new(:repositories, :relations, :mappers, :commands)
 
     # @return [Hash] configured repositories
     #
@@ -25,13 +27,49 @@ module ROM
     # @api public
     attr_reader :commands
 
+    # @return [Registry] mapper registry
+    #
+    # @api public
+    attr_reader :mappers
+
     # @api private
     def initialize(repositories, relations, readers, commands)
       @repositories = repositories
       @relations = relations
       @readers = readers
       @commands = commands
+      initialize_mappers
       freeze
+    end
+
+    # Load relation by name
+    #
+    # @example
+    #
+    #   rom.relation(:users)
+    #   rom.relation(:users) { |r| r.by_name('Jane') }
+    #
+    #   # with mapping
+    #   rom.relation(:users).map_with(:presenter)
+    #
+    #   rom.relation(:users) { |r| r.page(1) }.map_with(:presenter, :json_serializer)
+    #
+    # @param [Symbol] name of the relation to load
+    #
+    # @yield [Relation]
+    #
+    # @return [Relation::Loaded]
+    #
+    # @api public
+    def relation(name, &block)
+      tuples =
+        if block
+          yield(relations[name])
+        else
+          relations[name]
+        end
+
+      Relation::Loaded.new(tuples.to_a, mappers[name])
     end
 
     # Returns a reader with access to defined mappers
@@ -67,6 +105,16 @@ module ROM
     # @api public
     def command(name)
       commands[name]
+    end
+
+    private
+
+    # @api private
+    def initialize_mappers
+      elements = readers.each_with_object({}) { |(name, reader), h|
+        h[name] = reader.mappers
+      }
+      @mappers = Registry.new(elements)
     end
   end
 end
--- a/lib/rom/env.rb
+++ b/lib/rom/env.rb
@@ -69,7 +69,7 @@ module ROM
           relations[name]
         end
 
-      Relation::Loaded.new(tuples.to_a, mappers[name])
+      Relation::Loaded.new(tuples, mappers[name])
     end
 
     # Returns a reader with access to defined mappers
--- a/lib/rom/relation/loaded.rb
+++ b/lib/rom/relation/loaded.rb
@@ -1,0 +1,46 @@
+require 'rom/support/array_dataset'
+
+module ROM
+  class Relation
+    # Wraps loaded data from a relation and gives access to its mappers
+    #
+    # @public
+    class Loaded
+      include ArrayDataset
+
+      forward :first
+
+      # @return [ROM::MapperRegistry]
+      #
+      # @api private
+      attr_reader :mappers
+
+      # @api private
+      def initialize(data, mappers)
+        super
+        @mappers = mappers
+      end
+
+      # Map loaded relation using a specified mapper
+      #
+      # @example
+      #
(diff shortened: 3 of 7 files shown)


solnic:
Here's the "why" part of this PR explained in a short example:

``` ruby
require 'rom'
require 'virtus'

ROM.setup(:memory)

class User
  include Virtus.model
  attribute :first_name
  attribute :last_name
end

class UserPresenter < SimpleDelegator
  def full_name
    "#{first_name} #{last_name}"
  end
end

class Users < ROM::Relation[:memory]
end

class UserEntityMapper < ROM::Mapper
  relation :users

  model User
end

class UserPresenterMapper < UserEntityMapper
  register_as :presenter

  model UserPresenter
end

rom = ROM.finalize.env

rom.relations.users << { first_name: 'Jane', last_name: 'Doe' }

user = rom.relation(:users) { |r| r.to_a }.map_with(:users, :presenter).first

puts user.full_name
# Jane Doe
```

solnic:
I'd like to merge this in tomorrow so yell at me if you think it's not a good idea :)

maetl:
> Deprecate usage of Reader

Though I don’t have a problem with the reader conceptually, I think this change makes sense because the usage of `read` is subtly harder to document than `relation` (in rewriting the Rails tutorial, I struggled to find a good way to avoid the distinction between relation and reader which wasn’t important in that context). Not an issue for people who have used ROM a little bit, but for people getting started or trying to understand how it works, it’s an asymmetric concept that they touch on straight away.

The other thing is that `Relation::Loaded` helps explain something which I struggled with at first when building the Neo4j adapter—how and where relations are materialized.

solnic:
@maetl yeah, I think "reading defined relations" is a higher level concept that should be moved outside of ROM core and framework integration could leverage this feature somehow. I'm not sure about the implementation details yet.

I'm glad you pointed out explicit materialization via `rom.relation(:foo) {}` block, it's kind of unclear when it happens in the reader :)

solnic:
One thing that's lost here is `one` and `one!` interfaces. I should probably port them from `Reader`.

solnic:
I also wonder if `Relation::Loaded` should map relation via default mapper or should it always be explicit via `#map_with` interface. So ie:

``` ruby
# should this map by default using rom.mappers[:users]?
rom.relation(:users)

# ...probably not, so should we have a shortcut explicit interface instead? maybe:
rom.relation(:users).mapped # meh...

# ...maybe we should have a nice naming convention for mappers so that this would be possible:
rom.relation(:users).as_entities # does rom.mappers[:users][:entities].call(rom.relations[:users])
rom.relation(:users).as_presenters # does rom.mappers[:users][:presenters].call(rom.relations[:users])
```

Thoughts?

elskwid:
```ruby
user = rom.relation(:users).map_with(:entity, :presenter).first
```

I like this a lot. It is unsurprising and reads well.

As far as the default mapper, perhaps it's best we always use the explicit mapper until a common use-case emerges. I know you use rom way more than I do so perhaps the use-case is already there.

For specific feedback though, here are my thoughts:

```ruby
# should this map by default using rom.mappers[:users]?
rom.relation(:users)
```

Having `.relation(:users)` be mapped would be kind of surprising.

```ruby
# ...probably not, so should we have a shortcut explicit interface instead? maybe:
rom.relation(:users).mapped # meh...
```

I'm with you. Kind of 'meh' but it does make it clear what has happened.

```ruby
# ...maybe we should have a nice naming convention for mappers so that
# this would be possible:

# does rom.mappers[:users][:entities].call(rom.relations[:users])
rom.relation(:users).as_entities

 # does rom.mappers[:users][:presenters].call(rom.relations[:users])
rom.relation(:users).as_presenters

# it would of course work with a block
rom.relation(:users) { |r| r.by_name('Jane') }.as_entities
```

What about `with_presenters` / `with_entities`? Are these plural for another reason or is `entities` the name of a mapper? I would expect this to be `with_entity` or `with_presenter` based on the name of the defined mappers.

solnic:
@elskwid hmm good point, mappers could be registered under singular names, so how about:

``` ruby
rom.relation(:users).as(:presenters)
# or
rom.relation(:users).with(:presenters)
```

this way we'd avoid having to use meta-programming :) Not sure which one is more intention-revealing/clear to understand.

elskwid:
@solnic with that answer I think I see the idea. You are saying: "since I'm not providing a mapper name, use the default registered under `:presenter`". Does that sound right?

solnic:
Not really - :presenter is the mapper name registered for :users relation :)

elskwid:
WELP. Shows what I know!

muescha:
is the first example code snippet not too much boilerplate code? too many repetitions?

mjtko:
@muescha Can you expand on where you think there are too many repetitions?  I only see repetition between the attributes specified within the domain model and the mapper, and I don't think that's unexpected.

solnic:
@muescha you could skip the attributes part in the mapper and only specify models; we can also generate mappers from model definitions. I call it "convenience apis" and those are still not high priority (at least not for me). It is still unknown what are the common use cases for ROM, once we have more experience with real world usage various "convenience apis" will come out of that. I would say that's food for 2.0.0 of ROM. Personally I'm much more interested in building a really solid foundation for ROM in 1.0.0, get a felling how it's like to use it in production systems and plan 2.0.0 based on that experience. If we start adding a lot of convenience-oritented APIs too early it may clutter the codebases and make it harder to solidify foundation. It doesn't change the fact people can build add-ons that could make things simpler and DRY up code *already* (ie I have a spike with [rom-virtus](https://github.com/solnic/rom-virtus).

muescha:
@mjtko 

i commented out the repetitions:

```ruby
require 'rom'
require 'virtus'

ROM.setup(:memory)

class User
  include Virtus.model
  attribute :first_name
  attribute :last_name
end

class UserPresenter < SimpleDelegator
  def full_name
    "#{first_name} #{last_name}"
  end
end

# class Users < ROM::Relation[:memory]
# end
#
# class UserMapper < ROM::Mapper
#   relation :users
#
#   attribute :first_name
#   attribute :last_name
# end
#
# class UserEntityMapper < UserMapper
#   register_as :entity
#
#   model User
# end
#
# class UserPresenterMapper < UserEntityMapper
#   register_as :presenter
#
#   model UserPresenter
# end

rom = ROM.finalize.env

rom.relations.users << { first_name: 'Jane', last_name: 'Doe' }

user = rom.relation(:users).map_with(:entity, :presenter).first

puts user.full_name
# Jane Doe
```

but: maybe i miss something behind this example or what this example should show - because i am not so familar with rom

muescha:
should the `SimpleDelegator` not named as `SimpleDecorator`?

```ruby
class UserPresenter < SimpleDelegator
  def full_name
    "#{first_name} #{last_name}"
  end
end
```

mjtko:
Ah, yeah -- you've commented out some fairly critical pieces there!  You need the `Relation` to let ROM know where the data comes from and you need the `Mapper`(s) to let ROM know how to convert the data from the relation into domain models.

The example is showing a proposal for a new way of handling the Mappers which allows them to be used as a pipeline.

HTH!

solnic:
With some conventions in place we could configure entire environment from a minimum set of classes. As I mentioned that's doable and will be done at some point in the future. Right now we need to explicitly define mapper classes.

Re `SimpleDelegator` it's a class from ruby stdlib.

cflipse:
`SimpleDelegator` is a stdlib ruby class.

muescha:
ok thx @all for clarification

PR #157, opened 2015-02-16
Introduce spec specific namespace

splattael:
:warning: :construction:

As @solnic suggested in [Gitter](https://gitter.im/rom-rb/chat?at=54e1b3e01098fd39548a6bd7):

```
We should use a namespace for test constants
And clear only that namespace
```

Use  `ROMSpec` as namespace during specs.

Conclusion: It adds much noise to specs but makes cleaning spec env simple and straightforward.

Discuss!


solnic:
How about `module Test;end` in spec_helper and then just `Test::Something = Class.new` in specs??

solnic:
:clap:

PR #158, opened 2015-02-16
Hide inflections behind an internal interface

### LINKED ISSUE #156: Hide inflections behind an internal interface
Currently we have a hard dependency on `Inflecto` gem. ROM uses inflections in a couple of places and it would be nice to have internal interface with pluggable backends. Why? Because we could drop hard dependency on `Inflecto` and use a simple detection mechanism to see if there's any inflection backend available (in 90% ActiveSupport should be there heh) and just use that one. We could default to `Inflecto` when there's no AS.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8597071-hide-inflections-behind-an-internal-interface?utm_campaign=plugin&utm_content=tracker%2F155789&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F155789&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mjtko:
Drop hard dependency on Inflecto and use a simple detection mechanism to use ActiveSupport::Inflector if it's available. We still fall back to using Inflecto if not.

Fixes #156.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ group :test do
   gem 'virtus'
   gem 'minitest'
   gem 'thread_safe'
+  gem 'activesupport'
 
   platforms :rbx do
     gem 'rubysl-bigdecimal', platforms: :rbx
--- a/lib/rom.rb
+++ b/lib/rom.rb
@@ -1,10 +1,10 @@
 require 'equalizer'
-require 'inflecto'
 
 require 'rom/version'
 require 'rom/constants'
 
 # internal ROM support lib
+require 'rom/support/inflector'
 require 'rom/support/registry'
 require 'rom/support/options'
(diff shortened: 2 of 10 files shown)


mjtko:
Urgh, I have the benchmarks group installed locally so the ActiveRecord-related specs passed.  Two options as I see it:

1. remove the ActiveSupport-related specs
2. add `activesupport` to development dependencies in the gemspec

Thoughts?

solnic:
@mjtko we could add active_support to test group which includes dependencies used in tests

mjtko:
Ok, I've gone for adding it to the test group in the `Gemfile`.  Let's see what Travis has to say now! :)

PR #159, opened 2015-02-17
Remove `inflecto` from gemspec

mjtko:
The `inflecto` gem is no longer a dependency.  If neither `activesupport` nor `inflecto` are available, the inflector API raises a `LoadError` informing the user that they should install an inflector backend.

I'm not entirely happy with having to introduce the `setup` method to `ROM::Inflector` and call it immediately, but it was the only obvious way to make it testable.  Suggestions on patterns for cleaning that up welcomed. :smile:

### DIFF
--- a/lib/rom/support/inflector.rb
+++ b/lib/rom/support/inflector.rb
@@ -4,13 +4,19 @@ module ROM
   # @private
   module Inflector
     BACKENDS = {
-      activesupport: ['active_support/inflector', 'ActiveSupport::Inflector'],
-      inflecto: ['inflecto', 'Inflecto']
+      activesupport: [
+        'active_support/inflector',
+        proc { ::ActiveSupport::Inflector }
+      ],
+      inflecto: [
+        'inflecto',
+        proc { ::Inflecto }
+      ]
     }.freeze
 
-    def self.realize_backend(path, inflector_class)
+    def self.realize_backend(path, inflector_backend_factory)
       require path
-      Object.const_get(inflector_class)
+      inflector_backend_factory.call
     rescue LoadError
       nil
     end
--- a/lib/rom/support/inflector.rb
+++ b/lib/rom/support/inflector.rb
@@ -3,7 +3,7 @@ module ROM
   #
   # @private
   module Inflector
-    begin
+    def self.setup
       @inflector =
         begin
           require 'active_support/inflector'
@@ -13,8 +13,11 @@ module ROM
           ::Inflecto
         end
     rescue LoadError
-      raise 'Unable to find an inflector library'
+      raise LoadError,
+            "No inflector library could be found: "\
+            "please install either the `activesupport` or `inflecto` gem."
     end
+    setup
 
     def self.camelize(input)
       @inflector.camelize(input)
--- a/lib/rom/support/inflector.rb
+++ b/lib/rom/support/inflector.rb
@@ -4,8 +4,8 @@ module ROM
   # @private
   module Inflector
     BACKENDS = {
-      activesupport: ['active_support/inflector', '::ActiveSupport::Inflector'],
-      inflecto: ['inflecto', '::Inflecto']
+      activesupport: ['active_support/inflector', 'ActiveSupport::Inflector'],
+      inflecto: ['inflecto', 'Inflecto']
     }.freeze
 
     def self.realize_backend(path, inflector_class)
--- a/lib/rom/support/inflector.rb
+++ b/lib/rom/support/inflector.rb
@@ -3,48 +3,65 @@ module ROM
   #
   # @private
   module Inflector
-    def self.setup
-      @inflector =
-        begin
-          require 'active_support/inflector'
-          ::ActiveSupport::Inflector
-        rescue LoadError
-          require 'inflecto'
-          ::Inflecto
-        end
+    BACKENDS = {
+      activesupport: ['active_support/inflector', '::ActiveSupport::Inflector'],
+      inflecto: ['inflecto', '::Inflecto']
+    }.freeze
+
+    def self.realize_backend(path, inflector_class)
+      require path
+      Object.const_get(inflector_class)
     rescue LoadError
-      raise LoadError,
-            "No inflector library could be found: "\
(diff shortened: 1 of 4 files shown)


lib/rom/support/inflector.rb
```diff
@@ -13,8 +13,11 @@ module Inflector
           ::Inflecto
         end
     rescue LoadError
-      raise 'Unable to find an inflector library'
+      raise LoadError,
+            "No inflector library could be found: "\
+            "please install either the `activesupport` or `inflecto` gem."
```
solnic: :hand:  "please install either `inflecto` or `activesupport` gem."

spec/unit/rom/support/inflector_spec.rb
```diff
@@ -52,4 +56,27 @@
...
+            $LOAD_PATH.delete(g.lib_dirs_glob)
+          end
+        end
+        example.run
+      ensure
+        # restore the load path
+        $LOAD_PATH.clear.concat(original_load_path)
+        api.setup
+      end
+    end
```
solnic: In such special cases I would recommend our own interface for requiring a lib and stubbing it for the test purposes ie `expect(ROM::Inflector).to receive(:load_backend).and_raise(LoadError)` or something like that. This interface could also be tested by simply checking that after doing `ROM::Inflector.load_backend(:inflecto)` correct inflection backend is configured.

Does this make sense?

mjtko: Yeah, that sounds like a nicer approach.  I'll work something up, thanks!

lib/rom/support/inflector.rb
mjtko: Hehe, fair enough!  Do you think we should prioritise `inflecto` in the load ordering too?  I don't much mind either way! :smiley_cat:

solnic: Not really actually, if you're using Rails we want AS, you could have custom inflections configured etc.

spec/unit/rom/support/inflector_spec.rb
solnic: Let me know if you need help with this. It'd be awesome to include that patch in 0.6.0 :)

mjtko:
I've reworked things slightly - detection of the inflector backend is now not performed until the first time the inflector API is used which makes the code slightly neater at the cost of failing fast.  Tests are less strange now - we stub the `BACKENDS` constant to test what happens when a library cannot be found.

Of course, I'm open to suggestions on how to clean this up further.

spec/unit/rom/support/inflector_spec.rb
```diff
@@ -34,22 +34,68 @@
...
   context 'with detected inflector' do
-    it_behaves_like 'an inflector'
+    before do
+      if api.instance_variables.include?(:@inflector)
+        api.remove_instance_variable(:@inflector)
+      end
+    end
+
+    it 'prefers ActiveSupport::Inflector' do
+      expect(api.inflector == ::ActiveSupport::Inflector).to be true
```
solnic: I believe `expect(api.inflector).to be(ActiveSupport::Inflector)` is more idiomatic rspec

spec/unit/rom/support/inflector_spec.rb
```diff
@@ -34,22 +34,68 @@
...
+    before do
+      if api.instance_variables.include?(:@inflector)
+        api.remove_instance_variable(:@inflector)
+      end
+      stub_const("ROM::Inflector::BACKENDS", {})
+    end
+    it 'raises a LoadError' do
+      expect { api.inflector }.to raise_error(LoadError)
+    end
+  end
```
solnic: Hmm what's the intention of this test? I mean, `BACKENDS` is a super-duper private implementation detail, we should assume nothing will touch it.

lib/rom/support/inflector.rb
```diff
@@ -3,45 +3,65 @@ module ROM
...
-          ::Inflecto
-        end
+    BACKENDS = {
+      activesupport: ['active_support/inflector', '::ActiveSupport::Inflector'],
+      inflecto: ['inflecto', '::Inflecto']
+    }.freeze
+
+    def self.realize_backend(path, inflector_class)
+      require path
+      Object.const_get(inflector_class)
```
solnic: This doesn't work on jruby with namespaced constants ;(

```
irb(main):003:0> Object.const_get('::ActiveSupport::Inflector')
NameError: wrong constant name ::ActiveSupport::Inflector
```

solnic: ...you could wrap those in procs:

``` ruby
BACKENDS = {
  activesupport: ['active_support/inflector', proc { ActiveSupport::Inflector }],
  inflecto: ['inflecto', proc { Inflecto }]
}.freeze

require BACKENDS[:activesupport].first
BACKENDS[:activesupport].last.call # => ActiveSupport::Inflector
```

mjtko: That was there out of force of habit I guess!  I'll drop the leading `::`.  Seeing as we're going from `Object` that should be fine anyway.

spec/unit/rom/support/inflector_spec.rb
mjtko: Urgh, yup, not quite sure what I was thinking there.  Will sort that.

spec/unit/rom/support/inflector_spec.rb
mjtko: Just wanting to test that if neither activesupport or inflecto are available, we get an error raised.  We could bin or change this test to operate differently though.  Testing the result yielded by the presence/absence of libraries that are already present in the test environment is somewhat tricky.

Just drop it?

mjtko:
Sigh at jruby.  Ok, procs it is then! :-)

spec/unit/rom/support/inflector_spec.rb
solnic: Yeah that's tricky and not worth the effort IMO. I'd say we should drop this test.

mjtko: Yup, sure, fine by me.

PR #160, opened 2015-02-20
Lazy-relation API

solnic:
This adds a new API on top of `Relation` called `Relation::Lazy` which provides following features:

* ability to treat relations as callable "function" objects
* ability to compose two relations with left-to-right data pipelining
* ability to pass relations through mappers using the same left-to-right data pipelining
* lazy relations support auto-curry so you can do things like `user_by_id = users.by_id; user_by_id[1]`

This new API is used when you access relations via `rom.relation(:rel_name)`:

``` ruby
users = rom.relation(:users) # returns lazy-users

entity_users = users.map_with(:entities)
 # or
entity_users = users.as(:entites)

# then this works
entity_users.by_id(1)

# or partially-applied
entity_user_by_id = entity_users.by_id
entity_user_by_id[1]
```

### DIFF
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -1,3 +1,5 @@
+require 'rom/relation/lazy'
+
 module ROM
   # Base relation class
   #
@@ -187,6 +189,11 @@ module ROM
     # @api private
     def repository
       self.class.repository
+    end
+
+    # @api public
+    def to_lazy
+      Lazy.new(self)
     end
 
     private
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -1,4 +1,5 @@
 require 'rom/relation/lazy'
+require 'rom/relation/curried'
 
 module ROM
   # Base relation class
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -192,8 +192,8 @@ module ROM
     end
 
     # @api public
-    def to_lazy
-      Lazy.new(self)
+    def to_lazy(*args)
+      Lazy.new(self, *args)
     end
 
     private
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ rom.command(:users).create.call(name: "Joe", age: 17)
 rom.command(:users).create.call(name: "Jane", age: 18)
 
 # reading relations using defined mappers
-puts rom.relation(:users) { |r| r.by_name("Jane").adults }.as(:entities)
+puts rom.relation(:users) { |r| r.by_name("Jane").adults }.as(:entities).to_a.inspect
 # => [#<User:0x007fdba161cc48 @id=2, @name="Jane", @age=18>]
 ```
 
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ class Users < ROM::Relation[:memory]
   end
 
   def adults
-    find_all { |user| user[:age] >= 18 }
+    restrict { |user| user[:age] >= 18 }
   end
 end
 
@@ -87,7 +87,7 @@ rom = ROM.finalize.env
 
 # accessing defined commands
 rom.command(:users).create.call(name: "Joe", age: 17)
-rom.command(:users).create.call(name: "Joe", age: 17)
+rom.command(:users).create.call(name: "Jane", age: 18)
 
 # reading relations using defined mappers
 puts rom.relation(:users) { |r| r.by_name("Jane").adults }.as(:entities)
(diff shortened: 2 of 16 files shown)


lib/rom/relation.rb
```diff
@@ -189,6 +191,11 @@ def repository
       self.class.repository
     end
 
+    # @api public
+    def to_lazy(*args)
+      Lazy.new(self, *args)
+    end
```
cflipse: Naming convention from `Enumerable` suggests that a simple `#lazy` here would suffice.  Or at least as an alias.

cflipse:
Would it be possible to split this up a bit?

I'm finding the current `Relation::Loaded` structure to be quite awkward -- hard to build up and reuse things; it breaks the "drop to SQL" efficencies of aggregate functions, etc. Migrating code from `env.read` to `env.relation(){} ` feels like a step backwards in utility right now.  I'm fairly certain that the Lazy relations will address that; and I'd like to see that on Master soon.

Currying and composition are (extremely) nice to haves, but I don't think they should hold up the base case.

solnic:
@cflipse yeah I agree, I'll wrap it up today and mark auto-currying as uber-experimental and it'll be ready to merge

lib/rom/relation.rb
solnic: I decided on `to_lazy` because that's part of the relation interface which may conflict with app's code. ie somebody would like to define `Users#lazy` returning all the lazy people from the database :)

solnic:
@cflipse re `Relation::Loaded` - its mapping behavior becomes redundant now with `Relation::Lazy`. I'm gonna remove it (interface stays the same from the relation access pov, you can still do `rom.relation(:users) {}`)

lib/rom/relation.rb
cflipse: Ah,  okay, makes sense ... and since lazy is default return, user really isn't expected to call that.  And a relation is _not_ an Enumerable, I see.  Seems fair.  :)

solnic: Yep exactly :)

solnic:
OK I consider this as ready-to-merge. Please take a look and lemme know what you think.

gotar:
+1

cflipse:
Outside of `#first` returning a tuple, rather than using the mapper (one or two spec failures there) it seems well.  My pagination shim works, so long as it's the _last_ thing in the chain -- good enough for me, at the moment.

solnic:
@cflipse how do you use your pagination shim?

cflipse:
https://gist.github.com/cflipse/07beec78382b911ebd2b

solnic:
@cflipse so you just do `DatasetPagination.new(rom.relation(:foo).bar.baz))`??

cflipse:
Yep.

`DatasetPagination.new(rom.relation(:foo).bar.as(:foos)).page(2)`

solnic:
@cflipse you said "good enough" so I suspect you can imagine a nicer API for pagination, right? Any specific visions?

cflipse:
being less order dependant, for a start.

Right now, if I were to do `DatasetPagination.new(rom.relation(:foo)).bar.page(2)` it would fail, because the return from `bar` skips charlatan wrapping. 

This also means that, if I needed to do two such things, I'm pretty sure it would fail.

Also, for that matter, at present, I've still got to write a couple of methods in my relation.  They're not hard, by any means.

Maybe an include in the relation, but it would be more or less up to the adapter to define appropriate pagination methods.  Maybe being able to add strategies in a chain of responsiblity to the relation.

`rom.relation(:foo).paged.page(1).by_bar(223).as(:foos)`

The trick is, at the end of the day, that there's a need to be able to expose additional methods in the resulting object.  They don't have to be there by default, and a straight decorator can get very tricky to manage, but ...

The pagination methods are, in some part, basically just aggregate functions, asking for more data that we want to expose via the relation.  They're like `count` or `sum(:bar_count)` in that they're additional methods on the relation, that _don't_ return a relation.  And the implementation of them is going to be adapter specific.

cflipse:
`rom.relation(:foo).with(:pagination).with{|r| DataSummation.new(r) }.page(1).as(:foo_reports)`

spitballing  a different structure, maybe

solnic:
@cflipse right now the problem is that during data-pipelining the original relation gets lost in the process as mappers return arrays...*that* part is problematic :/ I want to experiment to see if we can maintain original relation in the loaded relation but it might be too complicated and error-prone. I just need to try out a couple of approaches and see if it's feasible.

cflipse:
ah!  So, it's not wonky on ordering _until_ I call `as`.  That's good to know; I thought I was misreading something in Charlatan

solnic:
@cflipse yeah `as` and `map_with` send relation through mappers so you end up with a mapped loaded relation and you no longer have access to methods defined on the relation

PR #177, opened 2015-03-13
Use Unsound::Control.try for CommandRegistry#try

pdswan:
A potential alternative to https://github.com/rom-rb/rom/pull/149/files.

I built a small [gem](https://github.com/pdswan/unsound) around the concept of `Either` and the versions of `try` implemented in `ROM::CommandRegistry` and in https://github.com/txus/kleisli in order to genericize. 

I considered just *using* kliesli (given some of the conceptual overlap I assume it's been evaluated before), but wasn't comfortable with the default global helpers and wasn't a huge fan of separating `Try` from `Either`.

Is this direction at all appealing? Happy to polish off any rough edges and add collaborators if it is.

Cheers!

### DIFF
--- a/lib/rom/command_registry.rb
+++ b/lib/rom/command_registry.rb
@@ -1,4 +1,5 @@
 require 'rom/commands/result'
+require 'unsound'
 
 module ROM
   # Specialized registry class for commands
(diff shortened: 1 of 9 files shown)


solnic:
I'm definitely for using a 3rd party gem for this. I'll check out Unsound!

solnic:
ps. I allowed myself to edit the PR description and link to unsound repo on github

aflatter:
A pragmatic implementation of `Option` and `Result` would definitely be useful. :)

solnic:
Shall we move the discussion to an issue in unsound repo and talk about its interface etc. there? Once we have a nice interface in place (we may already have it!) we could come back here and integrate it with ROM.

pdswan:
I'm certainly up for that.

pdswan:
@solnic @aflatter [opened for discussion](https://github.com/pdswan/unsound/issues/2)

Issue #185, opened 2015-03-24
Support returning entities from commands


AMHOL:
Currently when you call a command, such as create:

```ruby
ROM.command(:users).create.call(name: 'Test', email: 'test@test.com')
```

The return value is a hash, in most cases it would be useful to return an entity, or collection of entities, this should be relatively simple to implement, suggested API for configuration is:

```ruby
class CreateUser < ROM::Commands::Create[:memory]
  register_as :create
  relation :users
  result :one, as: :entity # The as: :entity option here
end
```

AMHOL:
Wondering about Virtus style syntax for this, i.e.

```ruby
class CreateUser < ROM::Commands::Create[:memory]
  register_as :create
  relation :users
  result Array[User]
end
class CreateUser < ROM::Commands::Create[:memory]
  register_as :create
  relation :users
  result User
end
```

solnic:
no, sorry this can't be done like that - first of all it requires model constants to be defined when commands are loaded, secondly it makes command now how to instantiate objects (which is not their responsibility), thirdly - I *really don't like that syntax anymore*, it can't be used consistently and requires more complexity, it's going away in virtus 2.0.

AMHOL:
Yeah, thought about it when I went to bed last night, realise why you mentioned some kind of DI container as this method relies on constructor mass-assignment.

solnic:
This is done in #187 - pls take a look, I'd like to merge it in :)

PR #187, opened 2015-03-26
Add support for mapping command results (refs #185)

solnic:
Basically making this work:

``` ruby
rom.command(:users).as(:entity).create[name: 'Jane']
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ end
 
 group :test do
   gem 'virtus'
+  gem 'anima'
   gem 'minitest'
   gem 'thread_safe'
   gem 'activesupport'
--- a/lib/rom/command_registry.rb
+++ b/lib/rom/command_registry.rb
@@ -7,6 +7,13 @@ module ROM
   class CommandRegistry < Registry
     include Commands
 
+    attr_reader :options
+
+    def initialize(elements, options = {})
+      super(elements)
+      @options = options
+    end
+
     # Try to execute a command in a block
     #
     # @yield [command] Passes command to the block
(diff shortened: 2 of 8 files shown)


AMHOL:
Nice, looks good

solnic:
@cflipse @aflatter @splattael @gotar @elskwid thoughts? makes sense?

aflatter:
Yes and no. This PR fits the way ROM currently works, but I'm not sure what I think about specifying the same mappers over and over again for each query or command that you do. I'd vote for merging though because I do stuff like `Foo.new(rom.command(:create).call(...))`.

solnic:
we can extend command class definitions so that you can specify which mapper should always be used; however, I'd like to try to get away without having that and simply do that in my code like that:

``` ruby
# excuse me lack of a nice DI :P
class SomeObjectThatPersistsUserDataAndReturnsEntity
  def self.build
    new(ROM.env.command(:users).as(:entity).create)
  end

  def initialize(command)
    @command = command
  end

  def call(input)
    command.call(input)
  end
end
```

In general...that's how ROM is intended to be used, same applies to relations.

gotar:
I like it +1 its a little strand you call mapper before create command but can live with it +1

solnic:
It's actually not strange, it's inline with relation mapping - you can grab command registry and every command that it returns will use the mapper vs setting mapper for each command separately. It's actually more convenient to use it like that.

solnic:
One more thing - we could consider doing the `result :one, as: :entity` which would set `mapper_name` (or something) internally and then `CommandRegistry` could look if this option is set and automatically set that mapper for a particular command.

gotar:
`result :one, as: :entity` - +1

PR #188, opened 2015-03-31
Removing complex inflector wrapping

krainboltgreene:
Feature: ROM needs inflection based string manipulation methods.

Hurdle: There are multiple inflection libraries that are in common use, specifically `Inflector` and `ActiveSupport::Inflector`.

The first iteration of this feature was a commit that allowed the user to load rom and on using the internal inflector API it attempted to load either `activesupport` or `inflecto` and then wrap those objects.

Solutions:
  1. (Original) Determine which of a few set libraries are available and then attempt to load the libraries in an arbitrary order.
  2. (Suggested) Allow the user (or some setup gem, like rom-rails) to specify the core inflection object to wrap.
  3. Write our own inflection library.

Notes:

  1. The current inflection implementation only has a set list of inflection methods. The more we use, the more this "module" grows (or shrinks).
  2. Any manual loading requires documentation that manual loading is required.
  3. Autoloading from a set list means either ignoring custom inflectors the user has or writing more interfaces to add it as the specific inflector, in which case why bother with the autoloader?
  4. With the setter strategy you have to sortive worry or at least concern yourself with load order. Currently the tests pass because nothing in the "compile step" of the gem tries to access the inflector. If this ever changes suddenly you have a chicken/egg problem and now need to consider either a default library (problematic) or a null interface (Unexpected results, but allows for warning messages).

Ultimately I would view this as a net positive if someone submitted this to hamster, since it removes an API the gem shouldn't really care about.


solnic:
Note: this PR was made because it turned out that current implementation doesn't work with opal.rb

...and now I'm wondering why on earth a require doesn't work in opal? Are we doing something odd?

krainboltgreene:
It's a dynamic require. Can't require something when everything's already been transpiled.

solnic:
The we just need to make sure ROM's inflector is configured when the lib is required

krainboltgreene:
I still think that the wrapper is way more complex than actually needed.

solnic:
Yep it should be simplified; probably removing its laziness should help

muescha:
how about a very basic default reflector (just add `s` like `user -> users`) 

and if some one need an advance reflector then he need to setup a reflection gem (or in case of rom-rails select the rails reflector)

solnic:
I plan to add proper ROM configuration API with support for policy-over-configuration where you can inject custom objects that handle certain things like inferring `dataset` name from a relation class name etc. This means you'll be able to use whatever you want there but I think from a pragmatic POV we should, by default, use an inflector backed by either inflecto or AS::Inflector for policies that require some inflection logic. I agree that current implementation is a bit too complex. I'd say we should clean it up once configuration API is ready with policies. This will allow us to inject name-inference policies using a pre-configured ROM's inflector that is configured by default and doesn't use a lazy-require mechanism.

krainboltgreene:
Totally agree.

solnic:
I'm gonna close this since we need #193 done first in order to make it prettier.

PR #195, opened 2015-04-03
Plugin registry

### LINKED ISSUE #192: Plugin system
Core components should be extendible through plugins. We already have a pagination plugin in rom-sql and it would be great to generalize the plugin infrastructure so that it's easy to extend ROM.

Here's an API idea:

``` ruby
# it should be possible to register plugins
ROM.plugins do
  # global
  register :publisher, ROM::Plugin::Publisher, type: :command

  # adapter-specific
  adapter :sql do
    register :pagination, ROM::SQL::Plugin::Pagination, type: :relation
  end
end

# then plugins become available
  use :publisher
end

class Users < ROM::Relation[:sql]
  use :pagination
end
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10378761-plugin-system?utm_campaign=plugin&utm_content=tracker%2F155789&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F155789&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cflipse:
wip for peeking over my shoulder; I'll probably rebase this heavily before it's actually ready to merge.

closes #192

### DIFF
--- a/spec/unit/rom/plugin_spec.rb
+++ b/spec/unit/rom/plugin_spec.rb
@@ -69,5 +69,29 @@ describe "ROM::PluginRegistry" do
   end
 
 
+  it "allows definition of adapter restricted plugins" do
+    Test::LazyPlugin = Module.new do
+      def self.included(mod)
+        mod.exposed_relations << :lazy?
+      end
+
+      def lazy?
+        true
+      end
+    end
+
+    setup.plugins do
+      adapter :memory do
+        register :lazy, Test::LazyPlugin, type: :relation
+      end
+    end
+
+    setup.relation(:users) do
+      use :lazy
+    end
+
+    expect(env.relation(:users)).to be_lazy
+  end
+
 
 end
--- a/spec/unit/rom/plugin_spec.rb
+++ b/spec/unit/rom/plugin_spec.rb
@@ -8,6 +8,7 @@ describe "ROM::PluginRegistry" do
 
   before do
     Test::CommandPlugin   = Module.new
+    Test::MapperPlugin    = Module.new
     Test::RelationPlugin  = Module.new do
       def self.included(mod)
         mod.exposed_relations << :plugged_in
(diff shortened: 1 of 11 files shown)


spec/unit/rom/plugin_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
+
+describe "ROM::PluginRegistry" do
+
+  subject(:env) { setup.finalize }
+
+  let(:setup) { ROM.setup(:memory) }
+
+  before do
+    GlobalExtension = Module.new
```
solnic: You can wrap it in `Test::GlobalExtension` so that our rspec callbacks will clean it from the runtime

cflipse:
Okay,  everything specified in #192  seems to be covered by this.

Cleanup, moving stuff around, and documentation still need to happen, but generally:  thoughts?

solnic:
Great work. I'll try to port rom-sql plugins to this.

cflipse:
Just remembered, forgot to note this earlier:

One thing I had to do that I don't like,  demoed here:
https://github.com/cflipse/rom/commit/d05afbefc967444b88730cdb132fc69af8a7a1ba#diff-e0b55668c593b9e34a8aad512e592a24R113

Unless I'm missing something, I cannot access the adapter type within the class definition, so it has to be redundantly declared in the `use` call.  Is there some alternative to that?

solnic:
@cflipse can you remind me what we're missing to finish this? Specifically in terms of the missing interfaces?

cflipse:
functionally?  Nothing.  It could go in as-is, I think.

I'm not thrilled with repeating the adapter specification, however.  It would be good if the DSL were to be able to know it was operating in the context of an `:sql` adapter, and check there automatically, rather than it being required to be specified in the `use` line.

essentially, the `use` implementation should be able to check an `adapter_type` and use that, rather than requiring it to be specified.  This is complicated somewhat by the fact that the adapter types are registered, rather than hard-coded.  There's some reverse-lookup that can be done, I just havn't pulled on that thread.

solnic:
@cflipse OK thanks. I'm gonna try to improve that and merge this in (: Sorry it's taking so long

solnic:
Merged. I'll continue in master. Cheers (:

PR #196, opened 2015-04-04
Fix CommandRegistry#respond_to?

hecrj:
The `0.6.1` release has broken the `respond_to?` method in the `CommandRegistry`.
This fixes it.


hecrj:
This becomes useful when you want to empty all the relations that have a `delete` command when testing, for example.

solnic:
Nice catch. Thank you!

PR #198, opened 2015-04-07
Add unwrap to mappers

aflatter:
Nested hashes can now be lifted to the root via `unwrap`:

```ruby
unwrap :user do
  attribute :user_email, from: :email
end
```

This is blocked until the corresponding PR in transproc lands.


Gemfile
```diff
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'transproc', github: 'aflatter/transproc', branch: 'unwrap'
```
solnic: I just released transproc 0.1.3 witch `unwrap` support so please remove this and update gemspec to specify transproc ~> 0.1, >= 0.1.3

aflatter:
I'm planning to add support for `unwrap(:user)` to lift all keys before merging. Currently either a hash or block needs to be passed.

aflatter:
@solnic re supporting `unwrap` without explicitely specifying all keys: Would I have to rely on `header` here (my adapter currently does not define a header) or would it be okay to add another case [here](https://github.com/rom-rb/rom/blob/master/lib/rom/mapper/attribute_dsl.rb#L189)? Currently, I get an error when passing a Symbol and no block because `attributes_from_hash` tries to iterate over the first argument.

solnic:
@aflatter that mapper dsl has access to all attributes so you can use that I suppose. It's a nested array like `[[:id], [:email, from: :user_email]]`.

aflatter:
@solnic but then attributes have to be defined somewhere while I'd like to rely on the "schema" used by my adapter.

solnic:
@aflatter we removed "schema" concept in 0.6.0 so I'm not sure how your adapter defines a schema. Now you need to define all attributes in your mappers.

solnic:
@aflatter can we merge this in for a good start? We can add inferring key names later or tweak transproc to not require those keys, or whatever else that's needed. This PR starts to gather dust and I'm pushing now to release 0.7.0 in May :)

aflatter:
@solnic sure, go ahead. :+1:

solnic:
This is merged :tada:

Issue #173, opened 2015-03-11
Better error message for `association_join` with a non-existant association name


aflatter:
A `NoMethodError` is raised when calling `association_join` with an invalid association name. ROM should throw a more descriptive error. The offending code is [here](https://github.com/rom-rb/rom-sql/blob/master/lib/rom/sql/relation/associations.rb#L51).

```ruby
class UsersRelation
  def with_tasks
    association_join(:tasks)
  end
end

# ...

rom.relation(:users).with_tasks
# => NoMethodError:
 #       undefined method `[]' for nil:NilClass
```

PR #210, opened 2015-04-24
Add ROM::NoAssociationError

pdswan:
in order to enable work toward https://github.com/rom-rb/rom/issues/173


solnic:
Hm are we sure we want to have that error defined here?

pdswan:
My assumption was that all adapters (sql, mongo, yaml, redis, etc) support associations so all adapters may raise this error. If that's the case I thought it made sense for the error to live with the other top-level errors, but maybe that's incorrect?

solnic:
The only place where we have "association" concept is rom-sql at the moment. Hence my question.

pdswan:
ok, cool. i'll move it over there.

PR #212, opened 2015-05-04
Eager loading

### LINKED ISSUE #208: Eager loading for associations
TBD

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11584674-eager-loading-for-associations?utm_campaign=plugin&utm_content=tracker%2F155789&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F155789&utm_medium=issues&utm_source=github).
</bountysource-plugin>

solnic:
This adds `Relation::Lazy#eager_load` interface backed by `Relation::Graph`.

### TODO

- [x] add `Relation::Lazy#eager_load` which sets up a relation graph
- [x] make it possible to eager-load multiple times which results in a nested structure
- [x] make sure curried relations behave properly
- [x] make sure rom-sql will be able to build efficient queries

closes #208

### DIFF
--- a/spec/unit/rom/relation/lazy/eager_load_spec.rb
+++ b/spec/unit/rom/relation/lazy/eager_load_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe 'Eager loading' do
+describe ROM::Relation::Lazy, '#eager_load' do
   include_context 'users and tasks'
 
   before do
--- a/lib/rom/relation/graph.rb
+++ b/lib/rom/relation/graph.rb
@@ -1,27 +1,91 @@
+require 'rom/relation/loaded'
 require 'rom/relation/composite'
 
 module ROM
   class Relation
+    # Load a relation with its associations
+    #
+    # @example
+    #   ROM.setup(:memory)
+    #
+    #   class Users < ROM::Relation[:memory]
+    #   end
+    #
+    #   class Tasks < ROM::Relation[:memory]
+    #     def for_users(users)
+    #       restrict(user: users.map { |user| user[:name] })
+    #     end
+    #   end
+    #
+    #   rom = ROM.finalize.env
+    #
+    #   rom.relations[:users] << { name: 'Jane' }
+    #   rom.relations[:tasks] << { user: 'Jane', title: 'Do something' }
+    #
+    #   rom.relation(:users).eager_load(rom.relation(:tasks).for_users)
+    #
     # @api public
     class Graph
-      attr_reader :root, :nodes
+      # Root aka parent relation
+      #
+      # @return [Relation::Lazy]
+      #
+      # @api private
+      attr_reader :root
 
+      # Child relation nodes
+      #
+      # @return [Array<Relation::Lazy>]
+      #
+      # @api private
+      attr_reader :nodes
+
+      # @api private
       def initialize(root, nodes)
         @root = root
         @nodes = nodes
       end
 
+      # Compose left-to-right data pipeline
+      #
+      # @example
+      #   users_and_tasks = rom.relation(:users)
+      #     .eager_load(rom.relation(:tasks).for_users)
+      #
+      #   users_and_tasks >> proc { |users, children|
+      #     tasks = children.first
+      #     # do stuff
+      #   }
+      #
+      # @param [#call] other The right-side processing object
+      #
+      # @return [Relation::Composite]
+      #
       # @api public
       def >>(other)
         Composite.new(self, other)
       end
 
+      # Coerce this relation graph to an array
+      #
+      # @return [Array]
+      #
+      # @api public
+      def to_a
+        call.to_a
+      end
+      alias_method :to_ary, :to_a
+
+      # Materialize this relation graph
+      #
+      # @return [Loaded]
+      #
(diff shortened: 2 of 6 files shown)


solnic:
So, this is it, eager-loading. Thoughts? /cc @cflipse @aflatter @splattael @gotar @maetl

solnic:
I should mention that this adds eager-loading only to relation API. Support for mapping eager-loaded relations will be added in a separate PR

gotar:
+1

spec/unit/rom/relation/lazy/eager_load_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    }
+    }
+  }
+
+  let(:map_user_with_tasks) {
+    proc { |users, children|
+      map_users[users, map_tasks[children.first]]
+    }
+  }
+
```
maetl: Just noting here that there’s quite a lot of setup logic, before we reach the first test expectation, which makes me wonder if the test is testing these `:let` blocks as much as the actual eager load API.

solnic: Yep this could be safely removed, I'm keeping it for now as a reference for the further work in transproc/mapper.

maetl:
This is excellent.

It’s surprisingly abstract, and I found it difficult to comment on or explain my ideas, so apologies if the following doesn’t make much sense.

Things I like:

- Neat and concise layering/composition that leverages existing behaviour
- Cohesive API symmetry/design patterns (sharing consistent method arguments and the left-right pipelining)
- It requires a tiny amount of code

API design considerations:

- Is `#eager_load` the right name in the high level API? It describes what it is rather than what it does, and requires a bit of context around ORM patterns or experience with ActiveRecord to understand what it means, which feels a bit out of place with the rest of the API. A simple verb like `#load`, `#fetch`, `#materialize`, `#embed`, `#combine`, etc, might be more expressive.
- Unlike the rest of the `ROM::Relation` interface, it’s not immediately obvious at a glance what structure the results are returned as (nested objects? flat tuples? callable proc?)
- There’s a bit of an inside/outside ambiguity with the relation style—`#eager_load` is only available on the `Lazy` relation instance, which may confuse people who are drawn to trying to encapsulate these kinds of operations behind methods in the relation declaration itself. Some guidance around what does and doesn’t go in the relation classes/DSL would be useful (something to consider for the docs I guess).

solnic:
Thanks @maetl for this awesome feedback. Let me try to reply:

> Is #eager_load the right name in the high level API? It describes what it is rather than what it does, and requires a bit of context around ORM patterns or experience with ActiveRecord to understand what it means, which feels a bit out of place with the rest of the API. A simple verb like #load, #fetch, #materialize, #embed, #combine, etc, might be more expressive.

I used `eager_load` because that's how it's called in other Ruby ORMs although method names are different (AR's `includes` Sequel's `eager`). I'm absolutely happy to rename it to something shorter and more descriptive. Maybe `combine` would be good since we are combining results from multiple relations.

> Unlike the rest of the ROM::Relation interface, it’s not immediately obvious at a glance what structure the results are returned as (nested objects? flat tuples? callable proc?)

It returns parent relation with children in nested arrays. ie: `[users, [tasks, [tags]]]`. So left is parent and right are children. Mappers will be able to collapse that into an aggregate.

When I say "it returns" I mean that by calling it you get that result. This is a common pattern used for all ROM objects - they respond to `call` and also relations are coercible to arrays.

> There’s a bit of an inside/outside ambiguity with the relation style—#eager_load is only available on the Lazy relation instance, which may confuse people who are drawn to trying to encapsulate these kinds of operations behind methods in the relation declaration itself. Some guidance around what does and doesn’t go in the relation classes/DSL would be useful (something to consider for the docs I guess).

Yes this is something that must be documented. `Lazy` and friends are interfaces for composing relations and sending data through data-mapping pipeline whereas "plain" relations are meant to be simple data-access encapsulation layer and nothing more.

cflipse:
I like that; the graph structure looks quite useful, as I was considering that something simliar is needed for commands (create a user, stream it into several task creations, for example).

`#eager_load` might do with some synonym poking ... I like `combine`,  possibly even `merge` though I'm not sure that's the right connotation.  `nest` might be overload with other mapper related methods and cause confusion, and `tree` is probably too implementation revealing.

Something to emphasize in the docs might be that the right side *is* a tree structure; and therefore eager_load can come in with multiple arguments ... I see it in the specs, but it's not called out in the YARD.

Additionally, it *might* be good to be able to expand `nodes` in some way ... maybe `Graph#eager_load` which returns a new graph at the same level, with the additional ags appended to the node list?  Allows for gradual construction, still.

solnic:
> Additionally, it might be good to be able to expand nodes in some way ... maybe Graph#eager_load which returns a new graph at the same level, with the additional ags appended to the node list? Allows for gradual construction, still.

Yep I was planning to add that later on, I want to tackle mapping part first though as it may influence Graph API (it probably won't, but who knows).

FYI: `compose` has got 3 thumbs-up already (:

Issue #202, opened 2015-04-08
Bug memory dataset: order on field(s) with nil value raise exception


wafcio:
I found bug in enumerable_dataset.rb (memory dataset) in sort_by (order) method. When one field is nil and you want call order on this column it will raise error.

Example: You have data: [{ name: 'Jane', age: 24 }, { name: 'Joe', age: nil }] and you want order(:age)

```
comparison of Array with Array failed
/.../rom-2874edfbe4f4/lib/rom/support/enumerable_dataset.rb:53:in `sort_by’
/.../rom-2874edfbe4f4/lib/rom/support/enumerable_dataset.rb:53:in `sort_by'
/.../rom-2874edfbe4f4/lib/rom/memory/dataset.rb:51:in `order'
```

solnic:
Shall we put nils before numbers?

wafcio:
based on csv exmaple:

wafcio:
based on csv example:
id;name;age
1;Jane;24
2;Joe;

We have nil age for Joe, of course we can put for example 0 or " " but it doesn't look good

solnic:
I'm just wondering what the order should be, should it be `nil, 0, 1` or `0, 1, nil` :) Some dbs allow you to configure that behavior.

wafcio:
good point, but i think we can nil treat like 0

cflipse:
not quite ... nil != 0, particularly if you include negative numbers in your sorting.

Nil values don't flow into the sort; the question is if they should lead or trail the sorted order.    Different RDBMS' have different opinions on if it goes first or last, so it's a judgement call, but they SHOULD NOT appear in the middle of the list.  Personally, I'd sort nil's to the end.

solnic:
I agree with @cflipse - anybody up to fixing it? I'm planning to release 0.7.0 next Monday. I'm done with my tasks and won't have much time to do anything else.

solnic:
Fixed via 74b74dc

PR #225, opened 2015-05-15
Allow ordering relations with empty fields

nepalez:
Concerning #202 

By default `nil` values placed at the end, as @cflipse supposed.
Also added the option `nils_first: true` that places them at the beginning of the list.

### DIFF
--- a/lib/rom/memory/dataset.rb
+++ b/lib/rom/memory/dataset.rb
@@ -57,9 +57,8 @@ module ROM
         options = names.last.is_a?(Hash) ? names.pop : {}
         place   = options[:nils_first] ? -1 : 1
         compare = ->(a, b) {
-          return  place if a.nil?
-          return -place if b.nil?
-          a <=> b
+          return a <=> b unless a.nil? ^ b.nil?
+          a.nil? ? place : -place
         }
 
         sort do |a, b|
--- a/lib/rom/memory/dataset.rb
+++ b/lib/rom/memory/dataset.rb
@@ -50,11 +50,12 @@ module ROM
       #   Names of fields to order tuples by
       #
       # @option [Boolean] :nils_first (false)
-      #   Whether nil values should be placed before other ones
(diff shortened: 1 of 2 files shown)


nepalez:
damn! need to change options to conform to older jruby

lib/rom/memory/dataset.rb
```diff
@@ -46,9 +46,25 @@ def project(*names)
...
       def order(*names)
-        sort_by { |tuple| tuple.values_at(*names) }
+        options = names.last.is_a?(Hash) ? names.pop : {}
+        place   = options[:nils_first] ? -1 : 1
+        compare = ->(a, b) {
+          return  place if a.nil?
+          return -place if b.nil?
+          a <=> b
+        }
+
```
cflipse: Instead of all this; consider partitioning by `nil?` and sorting the remainder normally; then (ap|pre)pend the nil values

nepalez: Cannot understand howto do it for ordering by multiple names. Need to think.

solnic:
Looks good to me, I believe this option is actually useful. Thoughts? /cc @cflipse

lib/rom/memory/dataset.rb
cflipse: For that matter, I think a spec for ordering by multiple keys is missing.

solnic: Yeah definitely needs a spec with more than one tuple with nils.

solnic: Sorry, lemme clarify - a spec that tests a case where there are more than one tuple with more than one attribute with nil :) Makes sense?

nepalez: Definitely makes. Thanks to own new guest, John, I've recognized that i've missed differencies between minor fields. Need to polish it off for some time

nepalez:
Will send another PR after merging commits into the one

PR #226, opened 2015-05-15
Allow ordering relations with empty fields

nepalez:
Merged all commits from #225
Related to #202

### DIFF
--- a/lib/rom/memory/dataset.rb
+++ b/lib/rom/memory/dataset.rb
@@ -46,9 +46,24 @@ module ROM
 
       # Sort a dataset
       #
+      # @param [Array<Symbol>] names
+      #   Names of fields to order tuples by
+      #
+      # @option [Boolean] :nils_first (false)
+      #   Whether `nil` values should be placed before others
+      #
       # @api public
       def order(*names)
-        sort_by { |tuple| tuple.values_at(*names) }
+        options = names.last.is_a?(Hash) ? names.pop : {}
+        place   = options[:nils_first] ? -1 : 1
+        compare = ->(a, b) {
+          return a <=> b unless a.nil? ^ b.nil?
+          a.nil? ? place : -place
+        }
+
+        sort do |a, b|
+          names.map { |n| compare.call a[n], b[n] }.detect { |r| r != 0 } || 0
+        end
       end
 
       # Insert tuple into a dataset
--- a/lib/rom/memory/dataset.rb
+++ b/lib/rom/memory/dataset.rb
@@ -53,16 +53,13 @@ module ROM
       #   Whether `nil` values should be placed before others
       #
       # @api public
-      def order(*names)
-        options = names.last.is_a?(Hash) ? names.pop : {}
-        place   = options[:nils_first] ? -1 : 1
-        compare = ->(a, b) {
-          return a <=> b unless a.nil? ^ b.nil?
-          a.nil? ? place : -place
-        }
+      def order(*fields)
+        nils_first = fields.pop[:nils_first] if fields.last.is_a?(Hash)
 
         sort do |a, b|
-          names.map { |n| compare.call a[n], b[n] }.detect { |r| r != 0 } || 0
+          fields # finds the first difference between selected fields of tuples
+            .map { |n| __compare a[n], b[n], nils_first }
+            .detect { |r| r != 0 } || 0
         end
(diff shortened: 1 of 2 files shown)


lib/rom/memory/dataset.rb
```diff
@@ -46,9 +46,24 @@ def project(*names)
...
+      # @option [Boolean] :nils_first (false)
+      #   Whether `nil` values should be placed before others
+      #
       # @api public
       def order(*names)
-        sort_by { |tuple| tuple.values_at(*names) }
+        options = names.last.is_a?(Hash) ? names.pop : {}
+        place   = options[:nils_first] ? -1 : 1
+        compare = ->(a, b) {
+          return a <=> b unless a.nil? ^ b.nil?
```
AMHOL: Does the exclusive or here mean nils will be ordered using the normal natural order sort when only one value is nil? I think the partition and append/prepend method would be much cleaner?

spec/unit/rom/memory/relation_spec.rb
```diff
@@ -44,19 +50,34 @@
...
+    end
+
+    it 'places nil before other values when required' do
+      expect(relation.order(:age, :email, nils_first: true).to_a).to eq([
+        { name: 'John'                                 },
+        { name: 'Jill',          email: 'jill@doe.org' },
+        { name: 'Jane', age: 10, email: 'jane@doe.org' },
+        { name: 'Jack', age: 11                        },
+        { name: 'Jade', age: 11, email: 'jade@doe.org' },
+        { name: 'Joe',  age: 12, email: 'joe@doe.org'  }
```
cflipse: I'd like to see a non-nil secondary value sort in these example, to make sure that that's getting ordered as well.

So, maybe another:  `{ name: 'Jaques', age: 11, email: 'jaques' }` thrown into the mix, to make sure that the secondary key sorts, beyond just partitioning on nil

lib/rom/memory/dataset.rb
```diff
@@ -67,6 +79,14 @@ def delete(tuple)
         data.delete(tuple)
         self
       end
+
+      private
+
+      # Compares two values, that are either comparable, or can be nils
+      def __compare(a, b, nils_first)
+        return a <=> b unless a.nil? ^ b.nil?
+        (nils_first ^ b.nil?) ? -1 : 1
```
cflipse: Nice use of xor.  :)

lib/rom/memory/dataset.rb
```diff
@@ -46,9 +46,21 @@ def project(*names)
...
       # @api public
-      def order(*names)
-        sort_by { |tuple| tuple.values_at(*names) }
+      def order(*fields)
+        nils_first = fields.pop[:nils_first] if fields.last.is_a?(Hash)
+
+        sort do |a, b|
+          fields # finds the first difference between selected fields of tuples
+            .map { |n| __compare a[n], b[n], nils_first }
+            .detect { |r| r != 0 } || 0
```
cflipse: may or may not be clearer:
```
.detect(&:nonzero?) || 0
```

cflipse: I was going to say that detect provides an if-none argument, but that argument is called, and I think that:

`.detect(->{0}, &:nonzero?)` probably fails a clarity check. :)

nepalez: `<=>` (and `__compare`) returns `nil` when `a[n]` and `b[n]` are incomparable (against my expectation that they should be as values from the same field). In that case the whole ordering should report incompatibility. If I applied `nil.nonzero?` it would raise `NoMethodError` instead.

Maybe this edge case should be covered too. Mea culpa.

And wow! I tried `.detect(0) { ... }` and couldn't understand why it failed. I sooo like `.detect(->{0}) { ... }`!

nepalez: One more interesting question. How `order` should behave in case some values cannot be ordered?
```ruby
[
  { name: "Alice",   age: 7 },
  { name: "Knave", age: '1@2'.to_c }
]
```
It could either raise, or remain data unsorted. This needs a decision, I think

solnic:
why did you close it?

nepalez:
By mistake. I've got confused with those requests I made (by by own mistake) to master. Instead in a minute I'll send them to different branches

PR #227, opened 2015-05-16
Reject keys by default

nepalez:


solnic:
Oops, seems like this also includes commit with `order` refactor in memory adapter :/

PR #228, opened 2015-05-16
Reject keys by default

nepalez:
This PR should be discussed due to performance overhead

### DIFF
--- a/lib/rom/header.rb
+++ b/lib/rom/header.rb
@@ -20,7 +20,7 @@ module ROM
     # @api private
     option :model, reader: true
 
-    option :reject_keys, reader: true, default: false
+    option :reject_keys, reader: true, default: true
 
     # @api private
     attr_reader :attributes
--- a/lib/rom/mapper.rb
+++ b/lib/rom/mapper.rb
@@ -12,7 +12,7 @@ module ROM
       :prefix, :prefix_separator, :inherit_header, :reject_keys
 
     inherit_header true
-    reject_keys false
+    reject_keys true
     prefix_separator '_'.freeze
 
     # @return [Object] transformer object built by a processor
--- a/spec/integration/commands/update_spec.rb
+++ b/spec/integration/commands/update_spec.rb
@@ -109,6 +109,9 @@ describe 'Commands / Update' do
         define(:users) do
           model user_model
           register_as :entity
+
+          attribute :name
+          attribute :email
         end
       end
 
--- a/spec/integration/multi_repo_spec.rb
+++ b/spec/integration/multi_repo_spec.rb
@@ -26,6 +26,8 @@ describe 'Using in-memory repositories for cross-repo access' do
 
     setup.mappers do
       define(:users_and_tasks) do
+        attribute :user_id
+        attribute :name
         group tasks: [:title]
       end
     end
--- a/spec/integration/relations/reading_spec.rb
+++ b/spec/integration/relations/reading_spec.rb
@@ -133,7 +133,10 @@ describe 'Reading relations' do
     end
 
     setup.mappers do
-      define(:users)
+      define(:users) do
+        attribute :name
+        attribute :email
+      end
     end
 
     rom = setup.finalize
--- a/spec/integration/setup_spec.rb
+++ b/spec/integration/setup_spec.rb
@@ -104,6 +104,7 @@ describe 'Setting up ROM' do
         mappers do
           define(:users) do
             model Test::User
+            attribute :name
           end
         end
       }
@@ -139,6 +140,7 @@ describe 'Setting up ROM' do
       ROM.mappers do
         define(:users) do
           model Test::User
+          attribute :name
         end
       end
 
--- a/spec/unit/rom/mapper_spec.rb
+++ b/spec/unit/rom/mapper_spec.rb
@@ -29,8 +29,15 @@ describe ROM::Mapper do
 
   describe '.registry' do
     it 'builds mapper class registry for base and virtual relations' do
-      users = Class.new(ROM::Mapper) { relation(:users) }
-      entity = Class.new(ROM::Mapper) { relation(:users); register_as(:entity) }
+      users  = Class.new(ROM::Mapper) {
+        relation :users
+        reject_keys false
+      }
+      entity = Class.new(ROM::Mapper) {
+        relation :users
+        register_as :entity
+        reject_keys false
+      }
       active = Class.new(users) { relation(:active) }
       admins = Class.new(users) { relation(:admins) }
       custom = Class.new(users) { register_as(:custom) }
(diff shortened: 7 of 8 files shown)


nepalez:
This PR should be discussed due to performance overhead

nepalez:
The main reason why I'd prefer default value `reject_keys true` is that it makes the mapper interface non-intiutive and adds some use cases, the end-up user should take into account.

Expectations
------------

Let the relation `users` contains 3 keys:

```ruby
users.first
# { id: 1, name: "jane", email: "jane@doo.com" }
```

As a user that asks for intuitive API, I'd like the rule: "Just define the model and its attributes".

```ruby
  register_as :entity
  relation :users

  model OpenStruct
  attribute :id
  attribute :name
end

user.as(:entity)
# #<OpenStruct id=1, name="jane">
```

Pretty intuitive, innit?

Reality
-------

With `reject_keys false` instead of the one intuitive interface (see above), I have to select from two *different* interfaces.

1. For the case the relation has no unusable attributes:

```ruby
class User
  model OpenStruct
end
```
The fist thing I should notice is the `attribute` here is not necessary at all.

Another one is the method `attribute` *semantics* differs: it needed not for defining model attribute, but for key manipulation (renaming etc.).

```ruby
class User
  model OpenStruct
  attribute :first_name, from: :name
end
```

2. For the case some attribute should be filtered I have to use quite a different interface:

```ruby
class User
  model OpenStruct
  reject_keys
  attribute :id
  attribute :name
end
```

I mean, minor change in external data (relation) causes big change in the (mapper) interface.
This "interface instability" (from the end-up user's point of view) is what I'd like to exclude with `reject_keys true` by default.

solnic:
Frankly this is just one piece of the whole mapper/model setup puzzle. Things we should consider/discuss involves:

* if attributes must be defined only in mappers then we need to integrate with libs like virtus or anima in order not to have to duplicate attribute names in 2 places (ie load attributes from a virtus attribute set). this could look like that:

``` ruby
class Task
  include Virtus.model(coerce: false) # no need to coerce

  attribute :id
  attribute :title
end

class User
  include Virtus.model(coerce: false) # no need to coerce

  attribute :id
  attribute :name
  attribute :email
  attribute :tasks
end

class TaskMapper < ROM::Mapper
  model Task # a plugin detects a virtus model and infers attributes
end

  model User # a plugin detects a virtus model and infers attributes

  # we override :tasks (that's brittle) or...
  combine :tasks, mapper: TaskMapper

  # ... we could infer that from virtus attribute but that would require meta info on
  # virtus side which would pollute model definitions with mapping-specific bits
  # ie `attribute :tasks, combine: true` meh
end
```

* start treating mappers as more lower-level interface hidden by ROM and infer entire mapping definitions from models automatically, this would require a ton of additional info on the model side though, with various ROM-specific extensions, it would pollute 3rd party libs and I'm not really sure it's a good idea. But needs a discussion nevertheless.

* stop using models and create tiny data capsules and simply map to them, I'm leaning towards this more and more, this would mean dropping all model libs like virtus or anima and map to minimalistic objects where the only difference between hashes is having attr readers. This would obviously be extremely opinionated but we could easily have that as a promoted and recommended feature. Nobody stops you from mapping to your classes though but if we decide to make auto-generated classes a recommendation then we won't have to spend time figuring out how to integrate with 3rd party model libs

nepalez:
I would vote for the last option as well.

The main reason is that the first two options back on domain classes that are external to ROM.
I'd prefer to avoid such a dependency and shift responcibility for instantiating models to domain entities.

From the other hand, I see ROM mapper has a "[sleeping potential](https://github.com/rom-rb/rom/blob/master/lib/rom/model_builder.rb#L31)" for expanding to further integration to models - in just the opposite direction.

But evei if the ROM will evolve to not mapping to models (and my examples above will sink into oblivion), I think we still should explore whether `reject_keys false` adds complexity of described sort to mapper interface.

It seems to me, that this option is not an ordinary feature, but has deeper impact on how users will perceive mappers:

* in case of `reject_keys true` the mapper is a tool to construct its output with a clean state, using relation-provided data as a source of raw data. The `attribute` becomes basic element for such a construction.

* in case of `reject_keys false` the mapper is a tool to adopt relation to the domain interface. I think in this case not an `attribute`, but a sort of a `filter` seems more natural. User needs not to include something to output, but exclude it from the source instead.

solnic:
We need more time to figure this stuff out. I actually like clean entity definitions that simply list attribute names and embedded object types. Mapper definitions are much more complex. This makes me think it's not a good idea to treat mapper as a source of knowledge about the application data-structures. Not to mention that mappers are not part of the domain layer, entities are. So I dunno, a consensus here would be to turn it on by default but have a way of inferring attribute names from entity classes like I mentioned. Maybe we should try to do it now. Performance impact won't be huge I believe.

As you can see I'm bouncing back and forth with this :)

cflipse:
I'm inclined to not want my domain /model layer dependent on _any_ non-model API externalities for its structure ... that includes ROM mappers, honestly. As I see it, mappers instructions for taking data from one place, and translating it into my domain models.

It's _possible_ (maybe not wise) to take data from several different sources and represent them using the same domain model; keeping the structural definitions in the mappers alone makes that harder.

I admit to only doing a shallow reading of this thread; I don't have a lot of focus this weekend.  My basic, least surprise idea would be that an empty mapper does nothing -- no key filters, no data translation, nothing gets dropped.  Any changes to the data have to be declaraed, so that I can read the mapper and know _exactly_ what it's doing.

Thus, I'd prefer to see `reject_keys` _false_ by default, and require the specific line to enable at.

solnic:
@cflipse yeah that's a good point, domain entity definitions should be separated and not concerned with mappers or anything rom-related

I also like your argument about knowing what a mapper does by just looking at it. Making mapper reject keys by default would be an implicit behavior. I'm not sure why some people made the assumption it would do that. Probably because that's how attribute definitions work in libraries like virtus or anima.

I really think now it's a matter of good documentation. There's a learning curve here and we won't be able to make ROM immediately obvious for everybody. What we can do is to minimize that learning curve of course but I believe we can't make everybody happy.

The interesting thing about using ROM is that you define your entities and value objects like you want them to look like and then implement relations and specify mappings so that you will get the expected result. I must admit it's such a significantly different workflow when compared to AR and documenting it is crucial.

maetl:
> I also like your argument about knowing what a mapper does by just looking at it. Making mapper reject keys by default would be an implicit behavior.

I also agree with this. I’ve always found the mappers to make sense intuitivel. I think of them as objects that represent a transformation, where the actual data structure morphing/transformations are declaratively specified. In my opinion, transformations like rejecting keys by default should be explicit rather than implicit, as that helps keep the actual implementation closer to the concept/role of the mapper classes.

solnic:
Yep I guess we're reaching a consensus here that it should be disabled by default, right?

nepalez:
Yes, I think so

solnic:
OK then, time to close this one. Thanks for your input!

PR #229, opened 2015-05-19
Allow to define block in mapper attribute

gotar:
In block you can overwrite attribute value, or modify it:

```ruby
attribute :foo do 'John' end
attribute :foo do |t| t.upcase end
attribute :foo, from: :bar do 1+2 end
```

### DIFF
--- a/lib/rom/processor/transproc.rb
+++ b/lib/rom/processor/transproc.rb
@@ -91,13 +91,17 @@ module ROM
 
       # Visit plain attribute
       #
+      # It will call block transformation if its used
+      #
       # If it's a typed attribute a coercion transformation is added
       #
       # @param [Header::Attribute] attribute
       #
       # @api private
       def visit_attribute(attribute)
-        if attribute.typed?
+        if attribute.meta[:coercer]
+          t(:map_value, attribute.name, attribute.meta[:coercer])
+        elsif attribute.typed?
           t(:map_value, attribute.name, t(:"to_#{attribute.type}"))
         end
       end
--- a/spec/integration/mappers/overwrite_attributes_value_spec.rb
+++ b/spec/integration/mappers/overwrite_attributes_value_spec.rb
@@ -1,0 +1,51 @@
+require 'spec_helper'
+require 'rom/memory'
+
+describe 'Mappers / Attributes value' do
+  let(:setup) { ROM.setup(:memory) }
+
+  before do
+    setup.relation(:users)
+  end
+
+  it 'allows to manipulate attribute value' do
+    class Test::UserMapper < ROM::Mapper
+      relation :users
+
+      attribute :id
+      attribute :name, from: :first_name do
+        'John'
+      end
+      attribute :age do
+        9+9
+      end
+      attribute :weight do |t|
+        t+15
+      end
+    end
+
+    rom = setup.finalize
+
+    rom.relations.users << {
+      id: 123,
+      first_name: 'Jane',
+      weight: 75
+    }
+
+    jane = rom.relation(:users).as(:users).first
+
+    expect(jane).to eql(id: 123, name: 'John', weight: 90, age: 18)
+  end
+
+  it 'raise ArgumentError if type and block used at the same time' do
+    expect {
+      class Test::UserMapper < ROM::Mapper
+        relation :users
+
+        attribute :name, type: :string do
+          'John'
+        end
+      end
+    }.to raise_error(ArgumentError, "can't specify type and block at the same time")
+  end
+end
--- a/lib/rom/mapper/attribute_dsl.rb
+++ b/lib/rom/mapper/attribute_dsl.rb
@@ -30,17 +30,22 @@ module ROM
         @reject_keys = options.fetch(:reject_keys)
       end
 
-      # Define a mapping attribute with its options
+      # Define a mapping attribute with its options and/or block
       #
       # @example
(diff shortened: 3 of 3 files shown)


lib/rom/processor/transproc.rb
```diff
@@ -99,6 +99,8 @@ def visit(attribute)
       def visit_attribute(attribute)
         if attribute.typed?
           t(:map_value, attribute.name, t(:"to_#{attribute.type}"))
+        elsif attribute.meta[:coercer]
+          t(:map_value, attribute.name, attribute.meta[:coercer])
         end
```
solnic: I believe `:coercer` should take precedence here and we should clearly document it, maybe even raise when somebody tries to define `type` and at the same time provide a coercion block.

spec/integration/mappers/overwrite_attributes_value_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  it 'allows to manipulate attribute value' do
+    class Test::UserMappe < ROM::Mapper
+      relation :users
+
+      attribute :id
+      attribute :name, from: :first_name do 'John' end
+      attribute :age do 9+9 end
+      attribute :weight do |t| t+15 end
```
solnic: please no inline `do .. end` :)

lib/rom/processor/transproc.rb
gotar: what error you think we should put there? ROM::Mapper::Error.new('You cannot specify type and use block in the same time!')?

solnic: the error should be raised in attribute dsl. ie `raise ArgumentError, "can't specify type and block at the same time" if options[:type] && block` or something like that

gotar:
```
diff --git a/lib/rom/processor/transproc.rb b/lib/rom/processor/transproc.rb
index 6dbd17d..9b6d35f 100644
--- a/lib/rom/processor/transproc.rb
+++ b/lib/rom/processor/transproc.rb
@@ -91,16 +91,21 @@ module ROM
 
       # Visit plain attribute
       #
+      # It will call block transformation if its used
+      #
       # If it's a typed attribute a coercion transformation is added
       #
+      # It will raise error if block and type used in the same time
+      #
       # @param [Header::Attribute] attribute
       #
       # @api private
       def visit_attribute(attribute)
-        if attribute.typed?
-          t(:map_value, attribute.name, t(:"to_#{attribute.type}"))
-        elsif attribute.meta[:coercer]
+        raise StandardError if attribute.meta[:coercer] && attribute.typed?
+        if attribute.meta[:coercer]
           t(:map_value, attribute.name, attribute.meta[:coercer])
+        elsif attribute.typed?
+          t(:map_value, attribute.name, t(:"to_#{attribute.type}"))
         end
       end
 ```

Comment (solnic) [inline lib/rom/processor/transproc.rb]: Nah man, error should be raised in `AttributeDSL#attribute` like I mentioned (: Here it's too low-level and too defensive I think.

File: spec/integration/mappers/overwrite_attributes_value_spec.rb (hunk 2 above)
Comment (gotar) [inline spec/integration/mappers/overwrite_attributes_value_spec.rb]: i try `{}` but it fails

Comment (gotar) [inline spec/integration/mappers/overwrite_attributes_value_spec.rb]: i can change to: 

```ruby
diff --git a/spec/integration/mappers/overwrite_attributes_value_spec.rb b/spec/integration/mappers/overwrite_attributes_value_spec.rb
index c453bf6..ab82c51 100644
--- a/spec/integration/mappers/overwrite_attributes_value_spec.rb
+++ b/spec/integration/mappers/overwrite_attributes_value_spec.rb
@@ -13,9 +13,15 @@ describe 'Mappers / Attributes value' do
       relation :users

       attribute :id
-      attribute :name, from: :first_name do 'John' end
-      attribute :age do 9+9 end
-      attribute :weight do |t| t+15 end
+      attribute :name, from: :first_name do
+        'John'
+      end
+      attribute :age do
+        9+9
+      end
+      attribute :weight do |t|
+        t+15
+      end
     end

     rom = setup.finalize
```

solnic: yes pls, rubocop would complain anyway

spec/integration/mappers/overwrite_attributes_value_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+describe 'Mappers / Attributes value' do
+  let(:setup) { ROM.setup(:memory) }
+
+  before do
+    setup.relation(:users)
+  end
+
+  it 'allows to manipulate attribute value' do
+    class Test::UserMappe < ROM::Mapper
```
solnic: missing `r` in mapper class name :)

gotar: done

solnic:
OK awesome work. Please merge once travis is green.

spec/integration/mappers/overwrite_attributes_value_spec.rb
cflipse: `attribute :foo { 1+1 }` fails because the braces bind much tighter than do/end -- ruby attempts to send the block to `:foo`, not to `attribute`

To work inline, you have to do `attribute(:foo) { 1 + 1 }`

PR #230, opened 2015-05-19
Fixed #order for memory relation

nepalez:
Allowed ordering by multiple keys when some values are empty

This is #226 selected to the separate branch

### DIFF
--- a/lib/rom/memory/dataset.rb
+++ b/lib/rom/memory/dataset.rb
@@ -51,9 +51,21 @@ module ROM
 
       # Sort a dataset
       #
+      # @param [Array<Symbol>] names
+      #   Names of fields to order tuples by
+      #
+      # @option [Boolean] :nils_first (false)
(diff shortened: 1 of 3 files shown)


lib/rom/memory/dataset.rb
```diff
@@ -67,6 +79,14 @@ def delete(tuple)
...
         self
       end
+
+      private
+
+      # Compares two values, that are either comparable, or can be nils
+      def __compare(a, b, nils_first)
+        return a <=> b unless a.nil? ^ b.nil?
+        (nils_first ^ b.nil?) ? -1 : 1
+      end
```
solnic: if we really need a private method please use `__compare__`.

nepalez: Changed

solnic:
This now doesn't merge because of a conflict with CHANGELOG + specs are failing :/

nepalez:
D'oh! Will see it

PR #231, opened 2015-05-19
Allow grouping attributes with the name of existing field:

nepalez:
This is a fix for the case a group has just the same name as one of its fields.

```ruby
class UsersMapper < ROM::Mapper
  group agenda: [:agenda]
end
```

Badly needs for somebody to review it!


solnic:
That's an interesting edge-case. The fix looks good. Merging in - thank you :)

solnic:
ps. in the future pls pls create topic branches in the canonical repository

PR #238, opened 2015-05-23
Support combine without a block

kwando:


solnic:
We need a test showing expected behavior :)

kwando:
I'll do it then =)

solnic:
yes please, in general no PR will be accepted without tests, even if a change is trivial in terms of implementation.

kwando:
Added a spec for the behavior.

solnic:
awesome @kwando :)

PR #245, opened 2015-05-28
Raise errors when configuring a mapper ambiguously

vrish88:
When I was perusing the docs I came across [these](https://github.com/rom-rb/rom-rb.org/blob/guides/source/doc-pages/guides/basics/mappers/wrapping.md#edge-cases-1):

> Don't define attributes inline along with the :mapper option! In this case the mapper won't be applied:

and 

> Don't use block along with the :mapper option! All definitions from inside the block will be ignored:

Instead of using documentation to let a developer know about a potential misuse I think it'd be better if the code told the dev. Plus this way it'll fail faster and the dev won't have figure out why his/her mapper isn't being used.

### DIFF
--- a/lib/rom/mapper/attribute_dsl.rb
+++ b/lib/rom/mapper/attribute_dsl.rb
@@ -163,6 +163,13 @@ module ROM
       #
       # @api public
       def wrap(*args, &block)
+        if args.first.is_a?(Hash) && block_given?
+          raise ROM::MapperMisconfigured, 'Cannot configure `wrap` using both options and a block'
+        end
+        if args.first.is_a?(Hash) && args.first[:mapper]
+          raise ROM::MapperMisconfigured, 'Cannot configure `wrap` using both options and a mapper'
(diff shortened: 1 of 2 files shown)


solnic:
Yes [all exceptions](https://github.com/rom-rb/rom/blob/master/lib/rom/constants.rb#L7-L13) end with `Error` word. Would be good to stay consistent.

vrish88:
@nepalez & @solnic anything else that needs to be addressed in this pr?

solnic:
Hmm one more thing - it would be the same situation with `group foo: [:bar, :baz]` no? This one should also raise if somebody tries to set `mapper` there.

vrish88:
@solnic @nepalez Ok, I finally got a chance to fix the implicit exception reference and I've applied the same check to `group`. In the process though I did a rebase from rom/master and needless to say I force pushed and lost some previous comments.

Anyways, lemme know what you think now.

PR #252, opened 2015-05-30
Update fold

nepalez:
Allow block syntax for `fold` operation. Just in the same way as for groups. Now:
```ruby
  fold :tasks do
    attribute :title
    attribute :priority
  end
```
does the same as:

```ruby
  fold tasks: [:title, :priority]
```
Also I understood, that I hasted with coupling of Fold and Group in the same attribute earlier. This was a bad abstraction. That's why I separated `Fold` attribute from the `Group` one.

This provided some duplication in code, that I'm going to deal with earlier, after all "nested" attributes will be done: `group`, `ungroup`, `fold`, `unfold`, `wrap`, `unwrap`. Then I'm planning to add some tests to cover their interaction and to do necessary refactoring.

Sounds like a plan, yes.

### DIFF
--- a/lib/rom/header.rb
+++ b/lib/rom/header.rb
@@ -107,6 +107,15 @@ module ROM
       by_type(Group)
     end
 
+    # Return all Fold attributes
+    #
+    # @return [Array<Fold>]
+    #
+    # @api private
+    def folds
+      by_type(Fold)
+    end
+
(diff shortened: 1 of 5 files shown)


solnic:
wow you really got into the mapper didn't you? :)

PR #258, opened 2015-06-01
Nested operations

nepalez:
Added `ungroup` and `unfold` as inversions for `group` and `fold`


solnic:
Thanks this is wonderful. Could you update changelog in master?

nepalez:
sure

PR #261, opened 2015-06-01
Use Types hash and get rid of massive if tree

AMHOL:


AMHOL:
Added `@private` to Types

nepalez:
:+1:

PR #264, opened 2015-06-02
Update prefixing

nepalez:
Make `prefix` and `prefix_separator` applicable inside the block …
Both operations affect attributes following them.

```ruby
  setup.mappers do
    define(:users) do
      prefix :user

      attribute :name
      wrap :contact do
        attribute :skype

        prefix :contact
        prefix_separator '.'
        attribute 'email'
      end
    end
  end

  users.first
  # => { user_name: 'Joe', user_skype: 'joe', 'contact.email' => 'joe@doe.org' }
  users.as(:users).first
  # => { user: 'Joe', contact: { skype: 'joe', 'email': 'joe@doe.org' } }
```
If the target key not symbolized, the prefixed source is treated as a string too

### DIFF
--- a/lib/rom/mapper/attribute_dsl.rb
+++ b/lib/rom/mapper/attribute_dsl.rb
@@ -14,8 +14,7 @@ module ROM
     class AttributeDSL
       include ModelDSL
 
-      attr_reader :attributes, :options, :symbolize_keys, :prefix,
-        :prefix_separator, :reject_keys
+      attr_reader :attributes, :options, :symbolize_keys, :reject_keys
 
       # @param [Array] attributes accumulator array
       # @param [Hash] options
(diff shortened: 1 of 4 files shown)


lib/rom/mapper/attribute_dsl.rb
```diff
@@ -30,6 +29,30 @@ def initialize(attributes, options)
...
+      #
+      # @example
+      #
+      #   dsl = AttributeDSL.new([])
+      #   dsl.attribute(:prefix, 'user')
+      #
+      # @api public
+      def prefix(value = nil)
+        @prefix = value
+      end
```
solnic: hmm I left a comment under a commit and it's not visible in this PR for some reason...so I'm gonna repeat myself:

this should be, conventionally, done like that:

``` ruby
def prefix(value = Undefined)
  if value.equal?(Undefined)
    @prefix
  else
    @prefix = value
  end
end
```

we could probably have that as a support extension and simply define those methods via some macro

solnic: ...oh and the advantage being that it works as a get-or-set method and would only change the value once

nepalez: Ah, I've got it. Will change both methods in a minute

Issue #263, opened 2015-06-02
Embedded doesn't support from


dekz:
We may be incorrectly using the the embedded syntax, but we can't use the `from:` option when passing in a mapper when using embedded (and some others).

```ruby
embedded :cats, from: 'CatDetails', mapper: CatMapper
```

We expected that syntax to work in conjunction with a CatMapper:
```ruby
class CatMapper
  model Cat
  attribute :name, from: 'sillyCateKeyName'
end
```

The following code seems to work, but we would prefer to pass mapper instead of always nesting in a block.

```ruby
    embedded :cats, from: 'CatDetails' do
      model Cat
      attribute :name, from: 'sillyCateKeyName'
    end
```

It just seems that the `from:` isn't being passed through in the embedded dsl method.

Is this intended usage, or have I completely missed something obvious?

dekz:
Ugh, silly me. Looks like you guys have already fixed this in f46e4f9e415c5606fc0e39c2474aeec83a912d16.

solnic:
Sorry you've been affected by this :) We're pushing 0.8.0 tomorrow (unless some unexpected blockers are found).

dekz:
No worries, cheers for that!

dekz:
I do have a quick question @solnic if you have a spare second. I'm trying to grok the usage of Mapping and how they relate to, well, Relations. 

I will try and use the examples from the tests that I have been seeing lately, and maybe once I figure it out I'll write up a quick wiki on this repo to help explain it to the next person.

Let's say I have a list like this:
```ruby
response = {
  task_list: 1,
  tasks: [
    { user: 'jacob', task_id: 1, title: 'do something' },
    { user: 'jacob', task_id: 2, title: 'do something else' },
] }
```

I want to map the data in a way that isn't a 1:1 translation of Hash into ruby models. Let's say this is a legacy mapping of data and translated into the Model of Users and Tasks.

```ruby
# Models
class TaskList
  attribute :id
  attribute :user
end

class User
  attribute :name
  attribute :tasks
end

class Task
  attribute :user
  attribute :title
  attribute :id
end
```

```ruby
class TaskListMapper < ROM::Mapper
  model TaskList
  register_as :task_list # Not 100% sure what this does as it isn't listed in the generators?

  attribute :id, from: :task_list
  embedded :user, from: :tasks do
    model User
    attribute :name, from: :user
    group :tasks, on: { name: :name } do
      attribute :title
    end
  end
end

```

[...]

PR #267, opened 2015-06-04
steps

dekz:
Have the ability to declare steps in a Mapper. This helps with the transformations are complex and not reusable.

```ruby
         step do
            prefix 'list'
            attribute :id
            unfold :tasks
          end

          step do
            unwrap :tasks do
              attribute :task_id
              attribute :name, from: :user
              attribute :task_title
            end
          end
```

#263

### DIFF
--- a/lib/rom/mapper.rb
+++ b/lib/rom/mapper.rb
@@ -61,12 +61,7 @@ module ROM
       else
 
         step_transformers = steps.map do |s|
-          mapper = new(Mapper.processors.fetch(processor).build(s.header), s.header)
-          if mapper.transformer.is_a?(Transproc::Function) || mapper.transformer.is_a?(Transproc::Composite)
-            mapper.transformer
-          else
-            Transproc(mapper.transformer)
-          end
+          new(Mapper.processors.fetch(processor).build(s.header), s.header).transformer
         end.inject(:+)
 
         new(step_transformers, header)
--- a/lib/rom/mapper.rb
+++ b/lib/rom/mapper.rb
@@ -1,12 +1,12 @@
 require 'rom/mapper/dsl'
 
 module ROM
-  # Mapper is a simple object that uses a transformer to load relations
+  # Mapper is a simple object that uses transformers to load relations
   #
   # @private
   class Mapper
     include DSL
-    include Equalizer.new(:transformer, :header)
+    include Equalizer.new(:transformers, :header)
 
     defines :relation, :register_as, :symbolize_keys,
       :prefix, :prefix_separator, :inherit_header, :reject_keys
@@ -15,12 +15,12 @@ module ROM
     reject_keys false
     prefix_separator '_'.freeze
 
-    # @return [Object] transformer object built by a processor
+    # @return [Object] transformers object built by a processor
     #
     # @api private
-    attr_reader :transformer
+    attr_reader :transformers
 
-    # @return [Header] header that was used to build the transformer
+    # @return [Header] header that was used to build the transformers
     #
     # @api private
     attr_reader :header
@@ -61,8 +61,8 @@ module ROM
       else
 
         step_transformers = steps.map do |s|
-          new(Mapper.processors.fetch(processor).build(s.header), s.header).transformer
-        end.inject(:+)
+          new(Mapper.processors.fetch(processor).build(s.header), s.header).transformers.first
+        end
 
         new(step_transformers, header)
       end
@@ -77,8 +77,8 @@ module ROM
     end
 
     # @api private
-    def initialize(transformer, header)
-      @transformer = transformer
+    def initialize(transformers, header)
+      @transformers = Array(transformers)
       @header = header
     end
 
@@ -89,11 +89,11 @@ module ROM
       header.model
     end
 
-    # Process a relation using the transformer
+    # Process a relation using the transformers
     #
     # @api private
     def call(relation)
-      transformer[relation.to_a]
+      transformers.reduce(relation.to_a) { |a, e| e.call(a) }
     end
   end
 end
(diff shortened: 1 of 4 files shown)


nepalez:
Ah. I've also started `step` locally.

lib/rom/mapper.rb
```diff
@@ -56,7 +56,21 @@ def self.register_processor(processor)
...
     # @api private
     def self.build(header = self.header, processor = :transproc)
-      new(Mapper.processors.fetch(processor).build(header), header)
+      if steps.empty?
+        new(Mapper.processors.fetch(processor).build(header), header)
+      else
+
+        step_transformers = steps.map do |s|
+          mapper = new(Mapper.processors.fetch(processor).build(s.header), s.header)
+          if mapper.transformer.is_a?(Transproc::Function) || mapper.transformer.is_a?(Transproc::Composite)
```
AMHOL: @dekz think we need to go down a different path with this, there shouldn't be any references to Transproc in this file, the idea is that you should be able to configure the processor that ROM uses.

dekz: Yeah good point, maybe I can make the assumption that transformers must have an interface for composition of `:+`.

lib/rom/mapper.rb
```diff
@@ -56,7 +56,16 @@ def self.register_processor(processor)
...
     # @api private
     def self.build(header = self.header, processor = :transproc)
-      new(Mapper.processors.fetch(processor).build(header), header)
+      if steps.empty?
+        new(Mapper.processors.fetch(processor).build(header), header)
+      else
+
+        step_transformers = steps.map do |s|
+          new(Mapper.processors.fetch(processor).build(s.header), s.header).transformer
+        end.inject(:+)
```
solnic: This assumes transformers respond to `:+` I wonder if we could avoid that somehow

solnic:
How about assuming a mapper always has an array of transformers and then:

``` ruby
def call(relation)
  transformers.reduce(relation.to_a) { |a, e| e.call(a) }
end
```

?

lib/rom/mapper/attribute_dsl.rb
```diff
@@ -94,6 +95,21 @@ def exclude(*names)
...
+      #
+      #   dsl.step do
+      #     attribute :name
+      #   end
+      #
+      # @api public
+      def step(options=EMPTY_HASH, &block)
+        raise(ArgumentError, "cannot mix outer attributes and steps") unless attributes.empty?
+        steps << new(options, &block)
+      end
```
solnic: Hmm *I think* this should belong to `Mapper::DSL::ClassMethods` in this case as it's a higher level concept than AttributeDSL stuff

nepalez:
What i'm doing is defining a new `Step` attribute and `visit_step` method that simply composes attributes from its header. I don't think we should define a new thing such as `steps` in `AttributeDSL`.

`step` is just the carrier of its header, nor more.

solnic:
@nepalez I feel like a Step attribute is an overkill, it complicates header and introduces a new requirement for processors if they want to support step syntax. If we can simply gather steps in an array and call them one-by-one things stay much simpler. WDYT?

nepalez:
Well, the complication of Header exists, of course. But the complication is the additional type of attribute, following the existing convention. And a `Processor::Transproc` receives one regular visitor method `visit_step` that also conventional.

Yes, we made processing one level deeper, but we should do this in any way.

tbh I haven't got a solution that doesn't break my specs. Let's discuss it a bit later.

nepalez:
Included this to guides

lib/rom/mapper.rb
```diff
@@ -1,12 +1,12 @@
 require 'rom/mapper/dsl'
 
 module ROM
-  # Mapper is a simple object that uses a transformer to load relations
+  # Mapper is a simple object that uses transformers to load relations
```
AMHOL: Optimus Prime

solnic: You have no idea how much I love this haha :)

lib/rom/mapper.rb
```diff
@@ -80,11 +91,11 @@ def model
       header.model
     end
 
-    # Process a relation using the transformer
+    # Process a relation using the transformers
     #
     # @api private
     def call(relation)
-      transformer[relation.to_a]
+      transformers.reduce(relation.to_a) { |a, e| e.call(a) }
```
AMHOL: Would use more descriptive variable names in the block here

nepalez: This one is rubocop-inspired
https://github.com/bbatsov/ruby-style-guide/issues/196

AMHOL: Fair enough

lib/rom/mapper.rb
```diff
@@ -56,7 +56,18 @@ def self.register_processor(processor)
...
-      new(Mapper.processors.fetch(processor).build(header), header)
+      processor = Mapper.processors.fetch(processor)
+
+      if steps.empty?
+        new(processor.build(header), header)
+      elsif attributes.any?
+        raise(MapperMisconfiguredError, "cannot mix outer attributes and steps")
+      else
+        step_transformers = steps.map(&:header).map { |step_header|
+          new(processor.build(step_header), step_header).transformers.first
```
AMHOL: Seems strange to new up an instance of self each time just to get its first transformer? Wouldn't `processor.build(step_header)` work here?

nepalez: Yes, it would be. I cannot change this PR, because it it the @dekz who owns it. Can only update after the merge

solnic:
:boom: :boom: :boom: :boom: :boom: :boom: :boom: :boom: :boom: :boom: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

solnic:
Thanks awesome work (:

PR #268, opened 2015-06-04
Support `exclude` attribute from root

nepalez:
Renamed old `exclude` to `remove` and made it private. Implemented the `exclude` as a special type of attribute that removes key.

This allows using `exclude` from the root, so now we can either whitelist (with `reject_keys`) or blacklist (with `exclude`) attributes.

```ruby
  relation :users
  register_as :users

  exclude :password
end
```


solnic:
Yeah this makes sense

PR #283, opened 2015-06-22
Fix memory's create/delete unit tests

errinlarsen:
The `unit/rom/memory/commands/create_spec.rb` and
`unit/rom/memory/commands/delete_spec.rb` were reversed;
`create_spec.rb` tested the Delete command and `delete_spec.rb` tested
the Create command.

Reverse the filenames.


solnic:
:tired_face:

PR #284, opened 2015-06-23
Transproc processor refactoring

martinciu:
This PR doesn't add any new features it's just refactoring making Transporc processor class "A" on code climate. I used it as a refactoring practice.

Few things:
- sometimes it doesn't do exactly the same (e.g. `visit_combined` method - but the specs passes :/
- I'm not sure about so many delegates on `Transproc::Attribute` class
- sometimes it is less explicit (e.g. `process_postprocessed` method)
- I'm not sure about introducing `CombineProcessor`, `AttributesProcessor`, `Preprocessor`. I wanted all processors' to accept `header` in the initializer. But on the other hand they seems to be not necessary addition.

Please let me know what you think about it.

Codeclimate analyze: https://codeclimate.com/github/martinciu/rom/compare/transproc-processor-refactoring#ratings

### DIFF
--- a/lib/rom/processor/transproc.rb
+++ b/lib/rom/processor/transproc.rb
@@ -3,6 +3,8 @@ require 'transproc/all'
 require 'rom/processor'
 
 require 'rom/processor/transproc/combine_processor'
+require 'rom/processor/transproc/attributes_processor'
+require 'rom/processor/transproc/rows_processor'
 
 module ROM
   class Processor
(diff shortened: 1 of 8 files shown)


solnic:
Sorry for not coming with any feedback for so long. `rom-mapper` was extracted already into a separate repo so I can't merge this here. But I gotta say I'm not sure if it's a good idea to refactor it now. We're still trying to figure out how the DSL should work exactly and also we need cleaner test coverage in `rom-mapper` before we can make a huge refactoring.

Let's figure things out in `rom-mapper` repo :) Thanks for the PR, I hope you at least got some insights into how that part of ROM works.

PR #288, opened 2015-07-17
ROM::Global => ROM::Environment (multi-env)

AMHOL:
This PR removes the auto-registration on inherited hooks and is a step towards removing the ROM global interface, allowing for multi-env and removing a lot of the global state. Auto-registration can still be achieved as @krisleech's `wisper` broadcasts `inherited` messages on the relevant class (`ROM::Relation`, `ROM::Mapper`, `ROM::Command`) when inherited.

```ruby
rom = ROM::Environment.new
rom.setup(:memory)

class Users < ROM::Memory::Relation
  dataset :users
end

  relation :users
  register_as :entity
  model Struct.new(:id, :name)
  attribute :id
  attribute :name
end

class CreateUser < ROM::Memory::Commands::Create
  relation :users
  register_as :create
  result :one
end

# Define relations/mappers/commands
rom.register_relation(Users)
rom.register_mapper(UserMapper)
rom.register_command(CreateUser)

memory1 = rom.finalize.env

rom = ROM::Environment.new
rom.setup(:memory)

# Define relations/mappers/commands

rom.register_relation(Users)
rom.register_mapper(UserMapper)
rom.register_command(CreateUser)

memory2 = rom.finalize.env

memory1.command(:users).create.call(id: 1, email: 'memory1@memory1.com')
memory2.command(:users).create.call(id: 2, email: 'memory2@memory2.com')
memory1.relation(:users).to_a
# => [{:id=>1, :email=>"memory1@memory1.com"}]
memory2.relation(:users).to_a
# => [{:id=>2, :email=>"memory2@memory2.com"}]
```
This gives much more freedom to the user as they can create multiple environments which will work completely independently of each other.

Currently the `ROM` module delegates to an instance `ROM::Environment` on `method_missing` for registering relations, mappers and commands - the idea is to add deprecation warnings when this happens to give people time to make the transition.

~~Without `ROM` delegating to the environment instance, `ROM::Relation[:adapter]` and friends will no longer work, as the adapter won't be registered, we could return a wrapper to lazily evaluate this when the class is registered. In addition to this, `ROM.register_adapter` will no longer be relevant, and it would be down to the user or integration to register the adapter, I like this approach as the user has the freedom to specify the identifier under which the adapter will be registered, but feel free to shout loudly if you disagree.~~

See also #289, which adds environment plugins, and an `auto_register` plugin which will register relations, mappers and commands with an environment that it is applied on as they are created.

### DIFF
--- a/lib/rom.rb
+++ b/lib/rom.rb
@@ -19,9 +19,8 @@ require 'rom/commands'
 # default mapper processor using Transproc gem
 require 'rom/processor/transproc'
 
-# support for global-style setup
-require 'rom/global'
-require 'rom/setup'
+# rom environments
+require 'rom/environment'
 
 # TODO: consider to make this part optional and don't require it here
 require 'rom/setup_dsl/setup'
@@ -30,9 +29,20 @@ require 'rom/setup_dsl/setup'
 require 'rom/env'
 
 module ROM
-  extend Global
+  @environment = ROM::Environment.new
 
-  RelationRegistry = Class.new(Registry)
+  class << self
+    def method_missing(method, *args, &block)
+      if @environment.respond_to?(method)
+        @environment.__send__(method, *args, &block)
+      else
+        super
+      end
+    end
+    def respond_to_missing?(method, _include_private = false)
+      @environment.respond_to?(method) || super
+    end
+  end
 end
 
 # register core plugins
--- a/lib/rom.rb
+++ b/lib/rom.rb
@@ -31,8 +31,8 @@ require 'rom/environment'
 # TODO: consider to make this part optional and don't require it here
 require 'rom/setup_dsl/setup'
 
-# env with registries
-require 'rom/env'
+# container with registries
+require 'rom/container'
 
 # register core plugins
 require 'rom/environment_plugins/auto_registration'
--- a/lib/rom.rb
+++ b/lib/rom.rb
@@ -13,6 +13,7 @@ require 'rom/support/guarded_inheritance_hook'
 require 'rom/support/inheritance_hook'
 
 # core parts
+require 'rom/environment_plugin'
 require 'rom/plugin'
 require 'rom/relation'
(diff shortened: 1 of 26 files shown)


AMHOL:
Been messing about with this and I think it kills the `ROM::Relation[:memory]` syntax entirely

AMHOL:
Alternative would be to keep the adapter and plugin registries global (on the ROM module) and move everything else, or deprecate and get everyone to move to `Users < ROM::Memory::Relation` syntax

AMHOL:
This now means people will need to call `ROM.use :auto_registration` to maintain current interface

solnic:
Trying to wrap my head around those changes. Overall looks like a really cool solution with the plugin.

So, one thing that looks weird is that we have now `Environment` which is mutable and then it ends up with `@env` being an instance of `Env`. Such environments.

I'm wondering how to name those things so it's less confusing. We definitely need an IM, mutable object for gathering components etc. but what we want to end up with is a frozen registry of instantiated components.

AMHOL:
Yeah, it doesn't really change much, just rather than `ROM` being a registry for `relations`, `commands` and `mappers`, `ROM::Environment` is now the registry, `ROM::Environment#finalize` and `ROM::Environment#env` just do the same as `ROM.finalize` and `ROM.env` before

solnic:
@AMHOL since `ROM::Environment` is the registry of components (aka relation, mapper and command classes) and `ROM::Env` is the registry of component **object** then I believe we definitely need to come up with better names for those :)

AMHOL:
I like `ROM::Environment` personally, it reads well, i.e.

```ruby
rom = ROM::Environment.new # make a new ROM environment
```

Agree that `ROM::Env` could change, sth like `ROM::GlobalRegistry` or `ROM::Container`?

solnic:
:+1: for `ROM::Container`

PR #295, opened 2015-08-07
Merge lazy into relation

solnic:
This makes **all relations lazy** and removes `Relation::Lazy` wrapper. It simplifies a lot as it removes one proxy from the stack. rom-repository benefits from this change too. Another cool thing is that `rom.relations[:name]` and `rom.relation(:name)` return the same thing, which used to confuse people.

Second thing is deprecation of `exposed_relations` setting which now has no effect. We can simply mark query DSLs as protected in adapters so that they won't leak to app layers.


maetl:
> `rom.relations[:name]` and `rom.relation(:name)` return the same thing

Thanks for resolving that.

AMHOL:
:+1: Looks good to me

PR #308, opened 2015-08-27
[WIP] Raise error when registering after finalization.

replaid:
I hit the Gitter room last night after spending some time confused about the silent dropping of registrations after `ROM.env.finalize` was called.

I poked in some code that I thought might raise an exception like @solnic described. It doesn’t seem to address my case (at least as I expressed it in my spec) and it does trigger a number of failures in other specs.

So please peek at this failing state of the code and provide guidance on where this could go. I’m happy to update this with improvements or a better approach. I can also update the other failing specs if it turns out that this is right and they’re wrong somehow.

### DIFF
--- a/lib/rom/environment.rb
+++ b/lib/rom/environment.rb
@@ -197,8 +197,23 @@ module ROM
       boot.mappers(*args, &block)
     end
 
+    # Null boot object installed upon finalization.
+    class FinalizedBoot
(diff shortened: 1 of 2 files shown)


solnic:
Since we're planning a major cleanup in setup/finalize/env code so let's address it there. I'll close this one for now.

Issue #266, opened 2015-06-03
ROM::Commands::Result::Success#success?


kwando:
I'd like to have #failure? and #success? methods available on Commands::Result

AMHOL:
:+1: I've thought about this before too

kwando:
It might not be very "functional" to have query those methods, but I often end up with "if result.error" code anyway

solnic:
This stuff ain't finished. We're looking at integration with Unsound gem from @pdswan

solnic:
OK let's add those methods. Any volunteers? :)

Snuff:
@solnic Hi! I can do it) I'm trying to use ROM in my project and I need use this feature:) I think I can realize it in 1 week.

Snuff:
Sorry for delay, I had a deadline on my work, I'm working on this issue.

solnic:
This was done in #310

PR #310, opened 2015-09-01
Add success? and failure? to ROM::Commands::Result

Snuff:
Hello @solnic, I realized #266 issue, please review my code :pray:

### DIFF
--- a/lib/rom/commands/result.rb
+++ b/lib/rom/commands/result.rb
@@ -24,6 +24,14 @@ module ROM
       end
       alias_method :to_a, :to_ary
 
+      def success?
+        !value.nil? && error.nil?
+      end
+
+      def failure?
+        value.nil? && !error.nil?
+      end
+
       # Success result has a value and no error
       #
       # @api public
--- a/lib/rom/commands/result.rb
+++ b/lib/rom/commands/result.rb
@@ -24,10 +24,16 @@ module ROM
       end
       alias_method :to_a, :to_ary
 
+      # Return true if command successful
+      #
+      # @api public
       def success?
         !value.nil? && error.nil?
       end
 
+      # Return true if command failed
+      #
+      # @api public
       def failure?
         value.nil? && !error.nil?
       end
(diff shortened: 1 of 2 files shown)


lib/rom/commands/result.rb
```diff
@@ -24,6 +24,20 @@ def to_ary
...
+      # @api public
+      def success?
+        !value.nil? && error.nil?
+      end
+
+      # Return true if command failed
+      #
+      # @api public
+      def failure?
+        value.nil? && !error.nil?
```
solnic: This could just become `! success?` :)

adz: Interestingly isn't the same... but possibly more correct. 
What if both are nil or both are present?

solnic: We should make sure that invalid state is not possible :)

adz: Good answer ;)

adz: So looks like we'll only ever instantiate "Success" or "Error" classes, so you could check EITHER @value OR @error and it would be sufficient. Although nothing stops someone going Result.new and having both nil, it's pretty unrealistic usage!

 ...it feels like a haskell GADT with a (not quite) unified object interface. In Haskell you would have a type that can be one or the other, and nothing else -- and you'd have to case-match on the type before being able to call "value" or "error".

WDYT about going type-based and on what you initialized rather than trying to infer:

    def success?
      is_a?(Success)
    end

I guess it depends on whether this is open for new classes to be built that inherit Result (oo style). Then this might be bad, but seems like this is going functional/type style (it's an Either).

Another alternative:

    class Success
      def success?
        true
      end
    ...
    class Error
      def success?
        false
      end
    ...

Snuff:
Hey @solnic, I refactored it. But we have error in jruby build. CI can't install gem `anima`, could you look?

solnic:
@Snuff we need to pin anima version to `0.2.0`, looks like latest version requires ruby >= 2.1.0 :(

solnic:
I merged it in, will take care of anima issue in master. Thanks for the PR :)

Issue #312, opened 2015-09-14
Graph updates do not filter


cflipse:
Graph commands work great for creates, or where the only filtering needed is based off the root element; however nested commands do not get applied well.  `Update` or `Delete` commands are applied to _all_ elements of a given relation.

Of additional interest is that all command `execute` methods will receive 2 arguments, where the standard delete command expects 0 arguments.  This probably points to the solution required -- making it possible to filter from within the `execute` command.

I suspect being able to do something like:

```
def execute(tuple, user)
  execute_on(relation.for_user(user).by_id(tuple[:id]))
end
```

would be a good solution,  but that essentially requires updating every adapter command to work correctly.

Additionally, I'm not sure that there's a way to cleanly separate update values, such that the update only happens to certain elements; this probably comes down to requiring a `super` loop in the exectute method, but I suspect that this is too easy to mess up.

```
  def execute(tuples,  user)
    for_user = relation.for_user(user)
    tuples.each do |tuple|
      super(execute_on(for_user.by_id(tuple[:id]), tuple))
    end
  end

  # avoids a situation where
  # [{ name: "Task One", id: 33 }, { name: "Task Two", id: 34 }] 
  # ends up naming all tasks "Task Two"
```

solnic:
So, if I understand this correctly, the problem is that we don't support restricting relations of update and delete commands when using a command graph.

The only API we have to restrict them is proxying to relation and returning a new command with restricted relation like `delete_user.by_id(1).call)` and there is no way to do `delete_user.call(:by_id, 1)`. I'm wondering if having that would help here somehow.

Regarding `update`, there is no API for updating many tuples where each tuple would require restricting a relation, an update is applied to the whole relation, when an update command receives many tuples I'm actually not sure what's gonna happen now. In case of a graph with an update command that receives many tuples we simply cannot support that. Changing `execute` logic to make it loop through tuples and make assumptions about how a relation should be restricted would be a bit against the design.

I feel like what we need is a way to configure graph to restrict commands and in case of many tuples it would loop through them and call a command separately. Maybe something like:

``` ruby
rom.command([
  :users, [{ update: :by_id, [:id] }]
])

# then graph internally would do something like:

input = { users: [ { id: 1, name: "Jane" }, { id: 2, name: "John" } ] }
input[:users].each { |tuple| rom.command(:users).by_id(tuple[:id]).call(tuple) }
```

solnic:
ps. actually my idea applies to both `update` and `delete` command types

[...]

PR #313, opened 2015-09-14
[REPRO] set up failing acceptance spec for update commands

cflipse:
This is a failing spec to demonstrate #312 

There are a couple of things that end up happening here;

Most notably, there's no particular way to actually filter the remove
commands, so the first failure case ends up removing _all_ tasks.
This can arguably be fixed from within the execute method of the remove
task, but most command bases don't provide an easy way to _internally_
modify the relation that gets used by the execute method.

There's a related filtering problem where updates are concerned, which
should be exercised and demonstrated by this spec as well, but the
specific failure is masked by the unfiltered delete.

### DIFF
--- a/lib/rom/commands/abstract.rb
+++ b/lib/rom/commands/abstract.rb
@@ -79,7 +79,7 @@ module ROM
       # @api public
       def curry(*args)
         if curry_args.empty? && args.first.is_a?(Graph::InputEvaluator)
-          Lazy.new(self, args.first)
+          Lazy.new(self, *args)
         else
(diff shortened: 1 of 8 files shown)


solnic:
I'm gonna wrap this up, a couple of use-cases are not supported yet and it needs way better spec coverage, I'll work on that and we should probably consider releasing 0.9.2 after it's merged.

Lemme know what you think

solnic:
I merged this in even though I remember you had some use-case that is still not covered but...it's already a step forward so it's valuable to have that part done already. Please report an issue describing other use-cases that you like to have supported.

PR #318, opened 2015-09-26
Rubocop automatic fixes

pvcarrera:


solnic:
Thanks :) I'm sure my vim would format a couple of things differently but it's fine ;)

PR #319, opened 2015-09-26
Kill 16 alive mutants @ ROM::Relation

dcarral:
I don't know if this is somehow useful (because as the deprecation warning explains, it will be removed in v1.0.0). I guess I chose a bad method to start killing mutants, switching to useful ones (in another branch) now ;)

Anyways, this PR reduces the number of alive mutants from 32 to 16 (@ ROM::Relation)


solnic:
Actually let's keep those tests, it's somewhat important to have that code working until we reach 1.0.0 final :)

PR #320, opened 2015-09-26
Kill several mutants @ Relation & Gateway

dcarral:
Kill several mutants by testing some methods in both Relation and Gateway classes.

Context: Exploring mutation testing during ROSSConf Berlin 2015 ;)


spec/unit/rom/gateway_spec.rb
```diff
@@ -82,9 +85,33 @@ class Gateway
...
+
+      expect { gateway.use_logger(params) }.to_not raise_error
+    end
+  end
+
+  describe '#extend_command_class' do
+    it 'returns the class sent as 1st parameter' do
+      klass = ROM::Relation
+
+      expect(gateway.extend_command_class(klass, "foo")).to eq(klass)
```
solnic: it would be nice to create an anonymous class and freeze it to make sure it simply returns the original one without changing it :)

dcarral:
Way better this way, thanks!

spec/unit/rom/relation_spec.rb
```diff
@@ -172,4 +172,19 @@ class Relation < ROM::Relation[:memory]
...
+      expect(relation.curried?).to be(false)
+    end
+  end
+
+  describe "#with" do
+    it "delegates to __new__, sending its dataset along the given options" do
+      opts = { :key => "value" }
+      allow(relation).to receive(:__new__).with(dataset, opts)
+
+      relation.with(opts)
```
solnic: This touches internals too much. We could have it like that instead:

``` ruby
new_relation = relation.with(key: 'value')

expect(new_relation.dataset).to be(relation.dataset)
expect(new_relation.options).to include(key: 'value')
```

dcarral:
Thanks for the feedback Piotr ;)

Fwiw, I had to use 'mappers' instead of 'key' since it was the single option allowed by the Relation model.

solnic:
w00t, we killed some mutants, thanks :)

PR #321, opened 2015-09-26
Use tr instead of gsub

pxlpnk:
According to this benchmark [1] tr is way faster compared to gsub
especially when just replacing a single character.

[1]: https://github.com/JuanitoFatas/fast-ruby#stringgsub-vs-stringtr-code


asaaki:
:heart_eyes_cat:

PR #322, opened 2015-10-01
Add adapter setting to Gateway

nepalez:
class ROM::CustomAdapter::Gateway < ROM::Gateway
      adapter :custom_adapter
    end

    ROM::CustomAdapter::Gateway.adapter     # => :custom_adapter
    ROM::CustomAdapter::Gateway.new.adapter # => :custom_adapter

Also added `adapter :memory` setting to memory adapter.

The `.adapter` is not required for backward-compatibility.
But if it hasn't been set, then calling `#adapter` (instance method) will
raise the exception.

    ROM::ImperfectAdapter::Gateway = Class.new(ROM::Gateway)
    ROM::ImperfectAdapter.new.adapter
    # => raises #<ROM::MissingAdapterIdentifierError ...>

== Added

* ROM::Gateway extends ROM::Support::ClassMacros
* ROM::Gateway.adapter (getter/setter)
* ROM::Gateway#adapter (getter only)

### DIFF
--- a/spec/unit/rom/gateway_spec.rb
+++ b/spec/unit/rom/gateway_spec.rb
@@ -87,4 +87,32 @@ describe ROM::Gateway do
       expect(gateway.disconnect).to be(nil)
     end
   end
+
+  describe '.adapter' do
+    let(:gateway_class) { Class.new(ROM::Gateway) }
+
+    it 'gets/sets the adapter' do
+      expect { gateway_class.adapter :custom }
+        .to change { gateway_class.adapter }
+        .from(nil)
+        .to(:custom)
+    end
+  end
+
+  describe "#adapter" do
+    before { Test::CustomGateway = Class.new(ROM::Gateway) }
+
+    let(:gateway) { Test::CustomGateway.new }
+
+    it 'returns class adapter' do
+      Test::CustomGateway.adapter :custom
+      expect(gateway.adapter).to eql :custom
+    end
+
+    it 'requires the adapter to be defined' do
+      expect { gateway.adapter }.to raise_error(
+        ROM::MissingAdapterIdentifierError, /Test::CustomGateway/
+      )
+    end
+  end # describe #adapter
 end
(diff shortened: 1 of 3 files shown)


spec/unit/rom/gateway_spec.rb
```diff
@@ -87,4 +87,38 @@ class Gateway
...
+        .to(:custom)
+    end
+  end
+
+  describe "#adapter" do
+    subject { ROM::CustomAdapter.new.adapter }
+
+    before do
+      ROM::CustomAdapter = Class.new(ROM::Gateway)
+    end
```
solnic: Please define it in the test namespace which gets cleared automatically:

``` ruby
class Test::CustomGateway < ROM::Gateway
  adapter :custom
end
```

...or just use an anonymous class instead

spec/unit/rom/gateway_spec.rb
```diff
@@ -87,4 +87,38 @@ class Gateway
...
+    it 'gets/sets the adapter' do
+      expect { gateway_class.adapter :custom }
+        .to change { gateway_class.adapter }
+        .from(nil)
+        .to(:custom)
+    end
+  end
+
+  describe "#adapter" do
+    subject { ROM::CustomAdapter.new.adapter }
```
solnic: Please don't use anonymous subjects like that, `subject` is meant to be used to set "object under test". Here you set it to a method call. This should be:

``` ruby
subject(:gateway) { ROM::CustomAdapter.new }

# and then
it 'requires the adapter to be defined' do
  expect { gateway.adapter }.to raise_error(ROM::MissingAdapterIdentifierError, /ROM::CustomAdapter/)
end
```

I know we've been writing a ton of specs with implicit subjects and inlined examples but it's history, those specs were difficult to maintain and hard to understand and follow for people.

nepalez: Ok, I'll correct this soon. (I just used to `subject` as "subject under test", but you're right - specs should be written in a consistent style)

PR #323, opened 2015-10-15
Add a simple DSL for building a command graph

solnic:
This adds a DSL for building up a command graph that can handle nested input:

``` ruby
# to handle { user: { name: "Jane" } }
rom.command do
  create(:users, from: :user)
end

# to handle { user: { name: "Jane", task { title: "Task" }  } }
rom.command do
  create(:users, from: :user) >> create(:tasks, from: :task)
end

# to handle { user: { name: "Jane", tasks: [{ title: "Task One" }, { title: "Task Two" }] } }
rom.command do
  create(:users, from: :user) >> create(:tasks)
end

# to handle:
# {
#   user: {
#     name: "Jane",
#     tasks: [
#       { title: "Task One", tags: [{ name: "Red"}] },
#       { title: "Task Two", tags: [{ name: "Blue" }] }
#     ]
#  }
# }
rom.command do
  create(:users, from: :user) >> (create(:tasks) >> create(:tags))
end

# to handle:
# {
#    user: {
#      name: "Jane",
#      tasks: [
#        { title: "Task One", tags: [{ name: "Red"}] },
#        { title: "Task Two", tags: [{ name: "Blue" }] }
#      ],
#      books: [
#        { title: "Foo", author: { name: "Jade" } },
#        { title: "Bar", author: { name: "Joe" } }
#     ]
#  }
# }

rom.command do
  create(:users, from: :user) >> (
    (create(:tasks) >> create(:tags)) +
    (create(:books) >> create(:authors, from: :author))
  )
end

# to handle input for an update:
# {
#   user: {
#     id: 1, name: "Jane Doe"
#   }
# }
#

rom.command do
  update(:users, from: :user) { |cmd, user| cmd.by_id(user[:id]) }
end
```


solnic:
@cflipse check this out, thoughts?

solnic:
Can I have some feedback on this? /cc @cflipse @AMHOL @nepalez @gotar @kwando @robinetmiller @endash

gotar:
+1 looks good for me, consistent and clean

nepalez:
Like it

endash:
Thumbs down on `instance_exec`. I know this is turning into a hobby horse of mine but using `instance_exec`/DSL-like method calling really shouldn't be encouraged if an explicit block parameter would suffice. The only appropriate use-case for using a DSL like this IMO is if the DSL operates "in its own country", not in regular ruby-land. As in, it should present as its own language and not ersatz-ruby. A decent example is configuring a table schema—not something you necessarily want/need explicit receivers for, but also not something that you'd do inline in your regular app code. I'm generally wary of unnecessary metaprogramming that monkeys with scope and whatnot. And if this syntax is useful as a "DSL" it's also useful as a regular API and could just be exposed that way for non-magic-practitioners.

More substantive feedback: I'm not sure if this clears up much confusion or merely swaps some tokens for others. The "+" for instance, at first blush I'm not sure what it's supposed to mean semantically in the API or what it translates into behind the scenes. The pipeline operator I know is used elsewhere, but if the point is an easy-to-use-and-read syntax, then semantically named methods might be a better way to go. I'll offer some suggestions on that, once I nail down what these commands are actually doing. (I'm in both a good and bad position to make these judgments because I know enough about command graphs to have a vague idea of what they're supposed to be capable of, but not enough to intuitively read these as a seasoned pro).

Part of the problem for me is just looking at the current hash-bashed commands and then looking at the samples neither one tells me anything about what they actually do. I'd have to open up the guides and puzzle through them one-by-one to figure that out in either case. @solnic could you add english-language descriptions of the resulting operation to help in that regard?

kwando:
@solnic I'd like something that communicates the structure of the input better. Something like this would be easy to read, not sure how feasible it would be to implement though.

```ruby
rom.command do
  create(:users){
    create(:tasks){
      create(:tags)
    }
    create(:books){
      create(:authors, from: :author)
    }
  }
end
````
@endash Apologies for the meta magic =P

**Comment by solnic** [comment]
@endash the intention here is to have something as concise as possible because defining a more complex graph can be messy. additional receiver would add clutter but if people don't mind that I can change that to yield a receiver.

As far as "english-language descriptions" go:

* `create(:users, from: :user) >> create(:tasks)` means a user tuple will be inserted into `:users` relation and the result will be passed to `create(:tasks)` command which will insert task tuples and associate them with the inserted user tuple. The final result is a user tuple with its task tuples

* `create(:users, from: :user) >> (create(:tasks) + create(:books))` means the same, except we're inserting tasks and books for the same user

@kwando I thought about syntax that would describe input structure but it's a bit hard :) I actually think current syntax does express the structure

I also thought about this:

``` ruby
create(:users, from: :user, create(:task), create(:books))
```

This would pretty much boil down to a single function call to build the graph

kwando:
@solnic Thanks for the clarifications ^^.  When I think about it, and with you clarifications above, you suggested API makes perfect sense.

I was thinking about the execution order of the commands, and thought about it backwards.. i.e. create task _then_ create user, but it ofc it has to be the other way around.

endash:
@solnic Well, the hash syntax *is* as concise as possible. Any sort of friendly API is necessarily going to add bulk, not remove it. The goal then should be clarity, rather than swapping method calls in for nested hashes. That said, we can't just add an explicit receiver to the current syntax designed without one in mind and then call it a day when it doesn't work well.

My proposal below structures the commands as a series of nested steps each with an explicit receiver, the result of the parent step. Pipelining is removed in favor of the more semantic block syntax. `+` is unnecessary, since the dev could just continue to call additional methods on the block param. Use of `method_missing` allows for natural language method names, and `each` makes it clear when a step is going to occur multiple times for each item in an array.

```
# { user: { name: "Jane"}}

rom.command do |c|
  c.create_user
end

# { admin: { name: "Jane"}}

rom.command do |c|
  c.create_user(:admin)
end

# { user: { name: "Jane", task: { title: "Task" }  } }

rom.command do |c|
  c.create_user do |user|
    user.create_task
  end
end

# { user: { name: "Jane", tasks: [{ title: "Task One" }, { title: "Task Two" }] } }

rom.command do |c|
  c.create_user do |user|
    user.create_tasks
  end
end

# {
#   user: {
#     name: "Jane",
#     tasks: [
#       { title: "Task One", tags: [{ name: "Red"}] },
#       { title: "Task Two", tags: [{ name: "Blue" }] }
#     ]
#  }
# }

rom.command do |c|
  c.create_user do |user|
    user.create_tasks.each do |task|
      task.create_tags
    end
  end
end

# {
#    user: {
#      name: "Jane",
#      tasks: [
#        { title: "Task One", tags: [{ name: "Red"}] },
#        { title: "Task Two", tags: [{ name: "Blue" }] }
#      ],
#      books: [
#        { title: "Foo", writteb_by: { name: "Jade" } },
#        { title: "Bar", written_by: { name: "Joe" } }
#     ]
#  }
# }

rom.command do |c|
  c.create_user do |user|
    user.create_tasks.each do |task|
      task.create_tags
    end
    user.create_books.each do |book|
      book.create_author(:written_by)
    end
  end
end
```

solnic:
@endash this looks nice, probably easier to grasp than my operator-based-kung-fu.

Now, who wants to implement it? :joy:

endash:
@solnic I can write the classes that will take my syntax and turn it into the current hash syntax, which should suffice for a first step.

solnic:
@endash if you can change current `Graph::DSL` in a way that its `call` method returns the expected structures everything else will just work, so like you said - it should suffice :)

endash:
@Solnic Can you verify these reference hashes for the current syntax:

(In the same order as in my previous comment)

```
# [{user: :users}]

# [{admin: :users}]

# [{user: :users, [:create, [:tasks, [:create]]]}]

# [{user: :users}, [:create, [:tasks. [:create, [:tags, [:create]]]]]]

# {user: :users}, [:create,[:tasks, [:create, [:tags, [:create]]]], [:books, [:create, [{written_by: :authors}, [:create]]]]]
```

solnic:
@endash looks good, one thing syntax-wise that should be different is `c.create_user` vs `c.users(:create)`, I think this way it will be simpler to implement and nicer, basically the receiver would use method name as the rel name and its arg as the command name, so `c.users(:users)` means `:users` relation and its `:create` command. Those names correspond to `register_as` values of relation/command. WDYT?

endash:
@solnic I would actually go with that, but flip it into `.create(:tags)`. The main reasons for me is that trying to determine the relation from either a singular or plural form is not trivial, so being explicit about relations is ideal.  If the API is providing a facade over the actual graph, then `create` as a method and the relation as an argument makes perfect sense, thinking in terms of processes rather than syntax trees. Plus, it reads naturally. Argument-before-method is definitely idiosyncratic and brings it back a bit towards just being a thin sugaring of `[:tags, [:create]]`, IMO.

For comparison, my original suggestion, your suggested change, and my amended suggestion in response:

```
create_user do |user|
  user.create_tasks.each do |task|
    task.create_tags
  end
  user.create_books.each do |book|
    book.create_author(:written_by)
  end
end

users(create: :user) do |user|
  user.tasks(:create).each do |task|
    task.tags(:create)
  end
  user.books(:create).each do |book|
    book.authors(create: :written_by)
  end
end

create(users: :user) do |user|
  user.create(:tasks).each do |task|
    task.create(:tags)
  end
  user.create(:books).each do |book|
    book.create(authors: :written_by)
  end
end
```

solnic:
@endash yeah this looks better :+1:

robinetmiller:
Late to the party, but FWIW, I completely agree with the last iteration there.  Speaking as a very inexperienced ROMmer, it's far easier for me to read and interpret. I think it's mostly due to being more ruby-ish and leveraging blocks for their strength - injecting logic to be run as part of the receiver. Having additional logic to run after is a perfect fit for that mechanism.  

I don't know what I'm adding here beyond reassurance that this is the best choice, so I'll stop here.

solnic:
@robinetmiller I agree, although I just realised there's a small drawback - command graph supports passing procs for update/delete commands, and now we use blocks for building up the graph, so I dunno how to tackle it /cc @endash

endash:
@solnic @robinetmiller I'm going to look at that more closely today. I've commented out the one test for procs, and made a note in the commit. To make sure I understand: the procs limit the scope, correct? So that 

```
      { user: :users },
      [
        { update: -> cmd, user { cmd.by_name(user[:name]) } },
```

The update gets applied to that subset of the relation?

solnic:
Correct. The block receives the input of a particular node and its parent
node input if it's present so that you can restrict command's relation

solnic:
@endash I guess we could have something like:

``` ruby
rom.command.update(:users, from: :user) do |user|
  user.update(tasks: -> cmd, user, task { cmd.by_user(user).by_id(tasks[:id]) }).each do |task|
    task.update(tags: -> cmd, task, tag { cmd.by_pk(task[:id], tag[:name]) })
  end
end
```

endash:
I was thinking something like...
```
rom.command.update(:users) do |user|
    user.update(:tasks).match_records { |relation, user, task| relation.by_user(user).by_id(task[:id]) }.each do |task|
        task.update(:tags).match_records { |relation, task, tag| relation.by_pk(task[:id], tag[:name] }
    end
end
````

**Comment by solnic** [comment]
Hmm maybe `s/match_records/restrict/` would be better?

**Comment by endash** [comment]
If we generalize the `Node` a bit more, we can use context to allow for something even more expressive:

```
rom.command.update(:users) do |user|
  tasks = user.tasks.restrict { |tasks, user, task| tasks.by_user(user).by_id(task[:id]) }
  user.update(tasks).each do |task|
    tags = task.tags.restrict { |tags, task, tag| tags.by_pk(task[:id], tag[:name]) }
    task.update(tags)
  end
end

It might make sense to swap the key/relation when the user specifies both. Instead of:

```
user.update(tasks: :jobs_is_a_key)
```

It would be:

```
user.update(jobs_is_a_key: :tasks)
```

Then we can do

```
  tags = task.tags.restrict { |tags, task, tag| tags.by_pk(task[:id], tag[:name]) }
  task.update(tags_key: tags)
```

for a restriction, or just

```
  task.update(tags_key: :tags)
```

for unrestricted.

solnic:
@endash this looks cool!

endash:
One drawback of the new proc syntax is performing a restriction at the root level:

```
    users = rom.command.restrict(:users, &users_proc)
    command = rom.command.update(user: users) do |user|
      tasks = user.restrict(:tasks, &tasks_proc)
      user.update(tasks)
    end
```

A possible refinement might allow something more like

```
    command = rom.command.restrict(:users, &users_proc).update do |user|
      tasks = user.restrict(:tasks, &tasks_proc)
      user.update(tasks)
    end
```

That would also work as a nested restriction

```
    command = rom.command.restrict(:users, &users_proc).update do |user|
      user.restrict(:tasks, &tasks_proc).update(from: :users_tasks)
    end
```

Note the explicit `from:` to avoid overloading the meaning of a standalone symbol being passed to `update`.

endash:
I've made those changes so that

```
    command = rom.command.restrict(:users, &users_proc).update do |user|
      user.restrict(:tasks) { |tasks, user, task| tasks.by_user_and_title(user[:name], task[:title]) }.update(from: :users_tasks)
    end
```

is now valid syntax.

solnic:
This looks good. I'll make a couple of minor tweaks later in master. Thanks for the work @endash :)

PR #326, opened 2015-10-27
WIP Dynamic gateways

sagmor:
Hi,

Posting this here to get some feedback since it's the first time I'm hacking on ROM :wink:

This PR adds `Relation#from` method which get's a relation and "re-target it" to a new gateway.

Current usage sample:
```ruby
ROM.use :auto_registration

ROM.setup(
  default: [:sql, 'postgres://master-server/default'],
  slave: [:sql, 'postgres://slave-server/default']
)

class Posts < ROM::Relation[:sql]
end

rom = ROM.finalize.env

# To query post's from :slave server
rom.relation(:posts).from(rom.gateways[:slave])

# Or from an arbitrary server
gateway = ROM::SQL::Gateway.new('postgres://some-other-server/database')
rom.relation(:posts).from(gateways)
```

It's pretty simple, yet it enables ROM to query from a database slave, to handle sharding, or to support a multi-database multi-tenant architecture

The problem I see right now is that this looses the current relation's scope (ie. if you do `relation.where(something).from(gateway)` the where gets lost)

And a cool feature could be to accept a symbol to fetch the gateway from the container to get something more like:
```
rom.relation(:posts).from(:slave)
```

### DIFF
--- a/lib/rom/relation.rb
+++ b/lib/rom/relation.rb
@@ -91,6 +91,15 @@ module ROM
       to_enum.to_a
     end
 
+    # Point a relation to a custom gateway
+    #
+    # @return A new relation targeting the new gateway
+    #
+    # @api public
+    def from(gateway)
+      __new__(gateway.dataset(self.class.name))
+    end
+
     # Return if this relation is curried
     #
     # @return [false]
--- a/spec/unit/rom/relation_spec.rb
+++ b/spec/unit/rom/relation_spec.rb
@@ -208,4 +208,12 @@ describe ROM::Relation do
       expect(new_relation.options).to include(custom_opts)
     end
   end
+
+  describe "#from" do
+    it "accept's a new gateway to fetch data from" do
+      gateway = double(dataset: ROM::Memory::Dataset.new([jane]) )
+
+      expect(relation.from(gateway).to_a).to eql([jane])
+    end
+  end
 end


solnic:
I'm wondering if we could introduce this on the dataset level and create an extension for relation that would allow us to tell the dataset which gateway should be used. A gateway with multi-db setup could return this special dataset by default and inject all other gateways to it (also configured). This of course would require some tweaks in the rom setup code because right now setup asks each gateway for datasets so they don't have access to all other gateways, but this can be changed to support this feature.

sagmor:
I'm liking the simplicity of this approach.

Right now I'm testing something like:
```ruby
class Tenant
  attr_reader :rom, :gateway

  def repository(name)
    rom.repository(name).from(gateway)
  end
end

tenant = Tenant.new(rom, gateway)

post_repo = PostRepository.new(tenant)
# And now every post is fetched from the defined gateway
```

Maybe just a `ROM::Container` proxy could do all the trick and leave everything else untouched.
For a slave configuration you could have something like:

```ruby
# ...
master = ROM.finalize.env
slave = Slave.new(master, options)

# Now, from the app's perspective they are interchangeable

# Fetch from slave
 PostRepository.new(slave)

# Or fetch from master
 PostRepository.new(master)
```

solnic:
@sagmor you're right, we could handle that easily from the `Container#relation` interface, we have all the things there but you will always get a vanilla dataset, so if that's acceptable it seems like we could easily support this feature like you suggest

sagmor:
Definitely acceptable :+1:

Wha't better? I add this on another gem? since it's kinda specific :smile: 
Name idea?

AMHOL:
Could add it as a plugin?

sagmor:
Ok, will play around with the idea on [sagmor/rom-sharding](https://github.com/sagmor/rom-sharding) and circle back if I get into something :wink:

PR #327, opened 2015-11-01
Refactor of Environment, Setup, etc.

endash:
## :warning:  Needs review :warning:

This is a complete refactoring of setup/environment/etc with major changes to the public setup API. A temporary `global_shim` file and `rom/support/loader` support the existing API for remaining global use cases. `rom-sql` and `rom-rails` require minor tweaks, which are addressed in separate pull requests.

### Tasks

- [x] Add additional unit tests for `Environment`, `Configurable`, `CreateContainer`, `...`
- [x] Refactor tests to avoid use of `global_shim` if possible (core tests are totally global free)
- [x] Refactor setup DSL completely out of core and into a plugin (`:macros`)
- [x] Refactor `rom-sql` tests to avoid use of `global_shim` if possible
- [x] Refactor `rom-rails` to avoid use of `global_shim` if possible

(rom-rails and rom-sql migrations rely on `rom/support/loader`)

### Overview of changes

`Environment` now only handles massaging the gateway configuration. `Setup` is completely gutted and simply registers/allows access to commands/mappers/relations during the bootstrapping process. `Configuration` is a fairly simple facade that mostly delegates to `Environment` or `Setup`, but is also now the target of environment plugins (now configuration plugins). `Finalize` is now used via the `CreateContainer` class (and the `#create_container` ROM module method). 

Global setup support (`ROM.setup`) is now quarantined in `rom/global_shim` and is NOT automatically required. Global state support(`ROM.env`) is quarantined in `rom/support/loader`, and is also NOT automatically required. `global_shim` uses `rom/support/loader`. `ROM.env` support is reliant upon container being created inside the `ROM::Support::Loader.load` block. 

"Local env" use revolves around `create_container`, which can work in a few ways:

-  By passing it a `Configuration`, it finalizes and returns a `Container`
-  By passing it an `Environment` and `Setup` it finalizes and returns a `Container`
-  By passing it setup info and a block, it will create a Configuration, call the block, and then finalize and return a `Container`

```ruby
configuration = ROM::Configuration.new(:memory)
ROM.create_container(configuration) # empty `Container`

configuration = ROM::Configuration.new(:memory)
configuration.register_relation(Users)
ROM.create_container(configuration) # `Container`

configuration = ROM::Configuration.new(:memory) do |config|
  config.register_relation(Users)
end
ROM.create_container(configuration) # `Container`

environment = ROM::Environment.new(:memory)
setup = ROM::Setup.new()
setup.register_relation(Users)
ROM.create_container(environment, configuration) # `Container`

ROM.create_container(:memory) do |config|
  config.register_relation(Users)
end # `Container`
```

Using `create_container`, in any form, does not automatically use `auto_registration`, nor does it give access to the DSL methods, currently (`Setup.mappers { define(:users) }`, etc). The DSL is exposed as the `:macros` plugin. `use()` can be called on a `Configuration` after creation or on the `Configuration` that is passed as the argument for the block form of `create_container`:

```ruby
ROM.create_container(:memory) do |config|
  config.use([:auto_registration, :macros])
end
```

### Diagram

Diagram


endash:
PR broken out into 4 commits which can be applied in sequence with passing tests each time: (1) basic refactor, minimal renaming/changes only as necessary outside of the core classes being touched (2) refactor SetupDSL to clean that up (3) Add the global shim/loader (4) the least essential change, making auto_registration retroactive now that `use(:auto_registration)` can't be called at require-time.

AMHOL:
Any chance we could remove references to Wisper and use something in rom-support like [this](https://github.com/rom-rb/rom-support/blob/master/lib/rom/support/publisher.rb)?

endash:
Hopefully we'll just remove most of it entirely, since it supports the legacy global stuff. I'm not sure if we have a final decision on that yet. /cc @solnic 

The auto_registration plugin uses `Wisper.subscribe` (entirely in theory) to dispose of subscribers. In practice, that'd only be useful in tests, to clean up that unfortunate bit of global state. In reality, the current release version of Wisper doesn't even support that yet so... yeah it could be removed (replaced by `klass.on`), especially if the globals stuff is cut out as well.

solnic:
Yeah I think we should either remove global stuff completely (along with wisper stuff) for 1.0.0. Frameworks can automate registration in a simpler way. The downside is that we won't be able to have a simple way of setting up ROM "standalone", it will require config, manual registration, and finalization. This makes me think we could introduce a simple "file loader" that would infer component class names from paths and auto-register them. It would be 10 x simpler than inheritance hooks and wisper. It's something that was already proposed months ago.

endash:
For now I've just removed the optional global support. Rom-sql will need another way to find a gateway to use for migrations, but that should be simple enough to do in its own gateway class. Rom-rails should be ok.

PR #328, opened 2015-11-04
Add `#migrator` to `Gateway`

nepalez:
The method instantiates and returns the migrator (if) provided by a corresponding adapter:

```ruby
setup = ROM.setup :cassandra, "127.0.0.1:9092"
# ...
setup.finalize
rom = setup.env
gateway = rom.gateways[:default]

gateway.migrator path: "db/migrate"
# => #<ROM::Cassandra::Migrator ...>
```

It is necessary for the `rom-migrator` plugin gem.

### DIFF
--- a/lib/rom/constants.rb
+++ b/lib/rom/constants.rb
@@ -11,6 +11,12 @@ module ROM
     end
   end
 
+  class MigratorNotPresentError < NotImplementedError
+    def initialize(adapter)
+      super "The #{adapter} adapter doesn't define a migrator"
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


endash:
It seems odd to have a separate gem for migrations but then to still have code in core that is specific to migrations support.

One alternative is an idea I was briefly kicking around this past weekend, which is to have a system-level plugin type. Then, `Gateway` `inherited` could be trapped just like `Relation`, et all, and have a module injected by the migrations plugin.

solnic:
Yes, this sounds like a use-case for a plugin interface.

PR #334, opened 2016-01-18
Make @directory a Pathname always

adz:
It appears to be expecting so, as '#dirname' is called on it further down.

### DIFF
--- a/lib/rom/setup/auto_registration.rb
+++ b/lib/rom/setup/auto_registration.rb
@@ -14,9 +14,9 @@ module ROM
 
     def initialize(directory, options = EMPTY_HASH)
       super
-      @directory = directory
+      @directory = Pathname(directory)
       @globs = Hash[[:relations, :commands, :mappers].map { |name|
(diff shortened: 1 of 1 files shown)


solnic:
no idea what's going on with travis build, but I merged it in nevertheless, let's see what'll happen on master build. thanks!

rpelyush:
Seems like this commit has broken ```AutoRegistration#initialize```. Now this line ```directory.join("#{name}/**/*.rb")``` produces the following error: ```NoMethodError: undefined method `join' for "":String```. I suppose ```@directory``` should be used in scope of #initialize not a ```directory``` argument.
```ruby
def initialize(directory, options = EMPTY_HASH)
      super
      @directory = Pathname(directory)
      @globs = Hash[[:relations, :commands, :mappers].map { |name|
        [name, @directory.join("#{name}/**/*.rb")]
      }]
    end
```

solnic:
@rpelyush just fixed it on master, thanks for pointing this out

PR #336, opened 2016-02-19
Enable DSL macros (relation, mappers, commands) by default and deprecate

endash:


endash:
Travis is erroring because of gem trubs. I just commented out the mutant stuff locally to get the specs to run. Not sure what to do as a universal solution.

solnic:
:+1:

PR #337, opened 2016-02-19
fix Container#disconnect calling each_key instead of each_value on ga…

endash:
…teways hash


solnic:
Lemme know if you need help with coming up with a test case for that fix.

PR #338, opened 2016-03-06
Make auto-registration use namespace value

tristil:
The auto-registration code was using the namespace value as a boolean
and actually inferring the target module name from the provided
directory name. This behavior doesn't seem intuitive (if desirable it
could be set through an additional option, like `infer_namespace: true`)
and contradicts the documentation on the website, which says the
namespace option is a string.

The referenced documentation is here: http://rom-rb.org/learn/advanced/flat-style/
image

### DIFF
--- a/lib/rom/setup/auto_registration.rb
+++ b/lib/rom/setup/auto_registration.rb
@@ -8,7 +8,7 @@ module ROM
   class AutoRegistration
     include Options
 
-    option :namespace, reader: true, type: [TrueClass, FalseClass], default: true
+    option :namespace, reader: true, type: [String, FalseClass], default: false
 
     attr_reader :globs, :directory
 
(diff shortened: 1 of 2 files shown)


tristil:
Actually, I see that this is just reverting a previous intentional change. I still think that the implementation that the doc example describes makes more sense. I have a project `lib/minder/database/{relations,commands,mappers}` where the top-level namespace is `Minder` and there is a `Minder::Database` namespace under that. With the current implementation if I point the path to `lib/minder/database` then it will look for `Database::Relations::Periods`, which fails.

solnic:
We should revisit this prior 2.0 release (scheduled for June). I'm gonna re-open.

lib/rom/setup/auto_registration.rb
```diff
@@ -44,7 +44,7 @@ def load_entities(entity)
     def const_name(entity, file)
       name =
         if namespace
-          file.gsub("#{directory.dirname}/", '')
+          "#{namespace}/#{entity}/#{File.basename(file, '.rb')}"
```
cflipse: I don't think this works well on windows; have to use `File.join` or something else that chooses the correct path separator

cflipse: Though, that's a note that's not isolated here, I think ...

solnic:
This has been finally addressed in #354

Issue #324, opened 2015-10-19
Improve error when command isn't defined 


robinetmiller:
```ruby
ROM.use :auto_registration
ROM.setup :memory

class Users < ROM::Relation[:memory]
  register_as :users
end

# when this is removed:
# class CreateUser < ROM::Commands::Create[:memory]
#   relation :users
#   register_as :create
# end

ROM.finalize

rom = ROM.env

create_user = rom.command(:users).create # ==> ElementNotFoundError
```

Produces `ROM::Registry::ElementNotFoundError: :users doesn't exist in ROM::Registry`, but that doesn't describe the precise problem, which is that there is no such *command*. A better error would be something like: 

`ROM::Registry::NoSuchCommandError: There is no :create command defined for :users relation.`

kwando:
+1 xD

kelvinst:
:+1:

PR #342, opened 2016-04-10
Improve error when command is not defined

thiagoa:
This pull request is part of the following issue: rom-rb/rom#324

I made the following changes:

- Removed `ROM::Registry` tests, which were moved over to the `rom-support` repo, in the following pull request: https://github.com/rom-rb/rom-support/pull/6
- Added tests for the new "default value" feature of `ROM::Registry#fetch`. The code is in `rom-support` (PR referenced above), but the tests are in this repository. To make this possible I temporarily pointed the `rom-support` gem reference to my `rom-support` fork, but this will get removed once the related pull request is OK.
- I'm using the "default value" feature to throw a specialized error in case a command is not found under a relation.


solnic:
Thanks for this. The specs should be moved to rom-support. Here we could only verify if command-specific registry has correct behavior.

thiagoa:
@solnic Thanks. I've just updated this pull request with the removal of the registry spec file, and the rom-support pull request was updated with the same spec file which was moved over from here.

Gemfile
```diff
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'rom-support', github: 'thiagoa/rom-support', branch: 'add-support-for-default-value-in-rom-registry-fetch'
```
solnic: This can be changed to `rom-rb/rom-support` and branch `master` now

solnic:
Could you rebase on top of master so that we can get a clean merge?

thiagoa:
@solnic All done

PR #343, opened 2016-04-18
Command #call supports a block

kwando:


AMHOL:
Nice one @kwando

solnic:
This expanded public API, please add a spec

PR #345, opened 2016-05-21
Support automatic schema inferring

flash-gordon:


flash-gordon:
Looking on Travis I wonder why bundler tries to install activerecord even if it was told not to install `benchmarks` section. Strange.

solnic:
I've noticed that too, maybe we're not using the right option?

flash-gordon:
I don't know, I guess it can be caused by using custom `script` option. Don't worry, I'll figure that out tonight :)

flash-gordon:
@solnic ahahah, we depends on activesupport [here](https://github.com/rom-rb/rom/blob/master/Gemfile#L18) ¯\\\_(ツ)_/¯
I'll try to remove it...

solnic:
@flash-gordon impossible, inflecto doesn't have dependencies

flash-gordon:
@solnic haha, 18th line was removed :) https://github.com/rom-rb/rom/commit/c66b6ecedf9122d386088120a7c5a544bede61df

solnic:
@flash-gordon ugh sorry /me disappears from github

PR #346, opened 2016-06-11
Replace thread_safe with concurrent-ruby in memory storage

flash-gordon:
Keep in mind that concurrent-ruby is a dry-t dependency.


solnic:
we should add a dependency on concurrent-ruby to gemspec

flash-gordon:
OK, done, I added the same version as in dry-t

PR #347, opened 2016-07-04
[WIP] Use a simple struct for relation name

flash-gordon:
This adds a simple struct object for a relation name which allows to use it more explicit in rom-sql and rom-repo. Instead of ad hoc names manipulations we are now able to use a single object that contains both dataset and relation names so that we can use one or another depending on the context. It is especially handful when dealing with curried relations in rom and relation proxies in rom-repo.

Current state: rough draft, no specs were written because this was a real challenge to make all specs pass in four (!) projects (rom, rom-sql, rom-repo, and my project which uses unusual DB naming). Things still don't work in all cases so I'm going fix this and add a bunch of specs to rom/rom-sql/rom-repo. First of all there is something wrong with combining associations in rom-repo though specs are passing...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ rvm:
   - rbx-2
   - jruby-9000
   - ruby-head
-  - jruby-head
 env:
   global:
     - JRUBY_OPTS='--dev -J-Xmx1024M'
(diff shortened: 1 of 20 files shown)


solnic:
OK I'm in "rom-rb mode" again. How can I help with this?

flash-gordon:
@solnic this now spread across four PRs (fresh one is https://github.com/rom-rb/rom-support/pull/7). And this works basically :) It's a weekend here so I'm going to add specs/docs today and besides it you're free to review these PRs so we could merge them into master one day :)
Apart from it, it would be great to add support for alias syntax for associations like `belongs :users, as: :author ` but it can be postponed until we get this merged.

solnic:
Awesome work man! Please also update CHANGELOG in master :)

flash-gordon:
Will do!

PR #350, opened 2016-08-10
Allow for configurable component_dirs in auto_registration setup

AMHOL:


solnic:
I thought this was addressing the namespace issue but it looks like it's a whole new feature :) I'm wondering though why and where do we need this?

AMHOL:
It was part of the problem @skrypalyk had with rom-rails yesterday, regarding already having a commands directory, plus I think it should be configurable anyway :p

AMHOL:
What's the namespace issue BTW?

solnic:
> It was part of the problem @skrypalyk had with rom-rails yesterday, regarding already having a commands directory, plus I think it should be configurable anyway :p

Oh so having that configurable would fix something in rom-rails? I'm not against this feature, just wanted to know where and how it'd be used.

> What's the namespace issue BTW?

dis one https://github.com/rom-rb/rom/pull/338

skrypalyk:
Thank you for doing this @AMHOL! I really appreciate it.

AMHOL:
NP, still needs an update in rom-rails after merge

PR #351, opened 2016-08-13
Allow a relation with a schema be in multiple containers

flash-gordon:
This fixes https://github.com/rom-rb/rom-sql/issues/78

### DIFF
--- a/lib/rom/schema.rb
+++ b/lib/rom/schema.rb
@@ -78,6 +78,8 @@ module ROM
     #
     # @api private
     def finalize!(gateway = nil, &block)
+      return self if frozen?
+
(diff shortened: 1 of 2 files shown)


flash-gordon:
@solnic is it fine?

lib/rom/schema.rb
```diff
@@ -83,5 +83,14 @@ def finalize!(gateway = nil, &block)
...
+
+    # Checks whether a schema was fully defined or not
+    #
+    # @return [Boolean]
+    #
+    # @api public
+    def finalized?
+      frozen?
+    end
   end
```
solnic: can't we just have `return self if frozen?` in `finalize!`?

flash-gordon: sure, done

Issue #353, opened 2016-09-06
Make auto_registration more flexible


solnic:
Currently `ROM::Configuration#auto_registration` supports `:namespace` option but it is only an on/off toggle, when it's turned on, then the first directory under provided path is treated as the namespace, then we assume there are `relations`, `commands` and `mappers` sub-dirs under that namespace dir. ie `auto_registration('lib/persistence', namespace: true)` will look for `lib/persistence/(relations|mappers|commands)/**/*.rb` and assumes module hierarchy is `Persistence::(Relations|Mappers|Commands::*`.

So, it'd be nice to make it more flexible, so that we can do this:

* `auto_registration('lib/persistence', namespace: true)` - current behavior is maintained
* `auto_registration('lib/persistence', namespace: false)` - current behavior is maintained
* `auto_registration('lib/persistence', namespace: 'MyApp::MyPersistence')` - a string can be set as the namespace, in which case the root dir name *is not going to be used as the namespace*. So it's gonna load `lib/persistence/(relations|mappers|commands)/**/*.rb` and assume that things are defined under `MyApp::MyPersistence` namespace ie `lib/persistence/relations/users` == `MyApp::MyPersistence::Relations::Users`

/cc @janjiss

Closes #338

solnic:
There's also #331 that wasn't finish, we could *probably* make it completely flexible by allowing passing a custom dir/constant resolver

janjiss:
@solnic Wondering about the api for custom dir/constant resolver. Do you think we should add another parameter or change existing one `:namespace` to allow custom resolver? If we use the same parameter name, can we have some kind of type that has to inherit from base strategy?

solnic:
Let's hold off with any further improvements until somebody asks for something more customized, then we can revisit.

wmaciejak:
Hi @solnic !
I wrote some months ago about my case on gitter. I don't like classic separation components by type (f.e. directories `app/relations/users.rb`, analogical `mappers` and `commands`) because IMO better, more efficient way is separation by business feature(f.e. `app/contexts/user/relation.rb`, `app/contexts/post/relation.rb`). I created one, quite big project with that approach and it saved my life(and common sense :D ). I used ROM v1, and created a custom registration of components cause that kind of directories configuration was unsupported.

I wonder if there is a way to smart auto-register components in that configuration or maybe it's so significant difference between approaches?

PR #354, opened 2016-09-07
Added ability to define custom namespace with String

janjiss:
Referencing issue #353 

I added ability to register namespaces via string and moved constant creation to strategies. Few questions:
* ~~Is it okay to use kw args or should I use Options module here?~~
* Should I move internal classes to separate files?
* Is the strategy approach okay here?

One benefit I see with strategies, that we can later pass our own to make it even more customisable as mentioned in #353 

Let me know what you thin and Ill be happy to adjust the code

**EDIT:**
I see that test suite expects Ruby 2.0, so changed the strategy classes to use ROM Options module

### DIFF
--- a/lib/rom/setup/auto_registration.rb
+++ b/lib/rom/setup/auto_registration.rb
@@ -6,7 +6,6 @@ require 'rom/support/options'
 
 module ROM
   class AutoRegistration
-    EXTENSION_REGEX = /\.rb$/.freeze
 
     include Options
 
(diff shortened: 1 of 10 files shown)


lib/rom/setup/auto_registration.rb
```diff
@@ -44,21 +44,67 @@ def mappers
...
-        Inflector.constantize(
-          const_name(component_dirs.fetch(entity), file)
+        klass_name = case
+        when namespace.class == String
+          CustomNamespaceStrategy.new(namespace: namespace, file: file).call
+        when namespace == true
+          WithNamespaceStrategy.new(file: file, directory: directory).call
+        when namespace == false
+          NoNamespaceStrategy.new(file: file, directory: directory, entity: component_dirs.fetch(entity)).call
+        end
```
solnic: This could simplified like that:

``` ruby
    case namespace
      when String
          CustomNamespaceStrategy.new(namespace: namespace, file: file).call
        when TrueClass
          WithNamespaceStrategy.new(file: file, directory: directory).call
        when FalseClass
          NoNamespaceStrategy.new(file: file, directory: directory, entity: component_dirs.fetch(entity)).call
        end
````

File: lib/rom/setup/auto_registration.rb [hunk 2]
```diff
@@ -44,21 +44,67 @@ def mappers
...
+      option :file, reader: true, type: String
+      option :entity, reader: true, type: Symbol
+
+      def call
+        Inflector.camelize(
+          file.sub(/^#{directory}\/#{entity}\//, '').sub(EXTENSION_REGEX, '')
+        )
+      end
+
+      private
```
solnic: Seems like we could have an abstract strategy to reduce some level of duplication ie:

``` ruby
class Strategy
  include Options
  option :file, reader: true, type: String
end
```

solnic:
> Should I move internal classes to separate files?

Yes please, esp when we add yard docs it's gonna get longer so a single file would be too big

> Is the strategy approach okay here?

It's perfect

lib/rom/setup/auto_registration.rb
```diff
@@ -10,7 +10,7 @@ class AutoRegistration
 
     include Options
 
-    option :namespace, reader: true, type: [TrueClass, FalseClass], default: true
+    option :namespace, reader: true, type: [TrueClass, FalseClass, String], default: true
```
solnic: This should be `allow: [...]` instead of `type:`

janjiss: Meaning that we want to change the api to accept `allow:` instead of `type:`?
Like so:
`allow: true`
`allow: false`
`allow: 'My::Namespace'`

?

janjiss: @solnic It seems not to work that way. 
```
   ROM::Options::InvalidOptionValueError:
        :namespace:true has incorrect value.
      # ./lib/rom/setup/auto_registration.rb:23:in `initialize'
      # ./lib/rom/setup.rb:49:in `new'
      # ./lib/rom/setup.rb:49:in `auto_registration'
      # ./spec/unit/rom/auto_registration_spec.rb:60:in `block (4 levels) in <top (required)>'
      # ./spec/support/constant_leak_finder.rb:6:in `find'
      # ./spec/spec_helper.rb:49:in `block (2 levels) in <top (required)>'
```

solnic: Oh wait of you're right. Sorry, this is what happens when I read code at midnight on a Wednesday 😆

janjiss:
@solnic Updated according to comments

PR #363, opened 2016-11-28
Add dry-core as a dependency

flash-gordon:
As a part of [leaving rom-support behind](https://github.com/dry-rb/dry-types/pull/132) I added dry-core to rom's dependencies and started to use its modules where possible.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,9 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'rom-support', git: 'https://github.com/rom-rb/rom-support.git', branch: 'do-not-load-constants'
+gem 'rom-mapper', git: 'https://github.com/rom-rb/rom-mapper.git', branch: 'dry-core'
+
 group :console do
   gem 'pry'
   gem 'pg', platforms: [:mri]
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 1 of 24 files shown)


solnic:
I wonder if this needs a 3.0.0 bump

flash-gordon:
Dunno, if you do not require `rom/support/constants` it's more or less backward compatible

solnic:
Sorry, I wasn't specific enough - I was wondering if we're gonna need a 3.0.0 bump if we kill rom-support *but* I just checked and all APIs that will be gone are private internal stuff used by rom, so we're good. I hardly imagine anyone using stuff like ClassBuilder etc. because it happens that rom provided that via rom-support. Other stuff (AutoCurry, EnumerableDataset and Options) will be moved to rom core so it won't change at all.

flash-gordon:
Keep in mind that _at this point_ we didn't touch rom-support ([the only exception](https://github.com/rom-rb/rom-support/commit/673256b41457b1fd8cb506bd7b59108910c688db)), so anyone is free to use it, i.e. he/she can `require 'rom/support/class_builder'` or something, and it won't affect ROM itself.

solnic:
Very good point :) Are you planning to move stuff from support to core too?

flash-gordon:
Exactly, this is the next part. I'm going to work on this on the next week (I fell off my motorbike and got some spare time lol, nothing serious, just won't swim for a week or so).

solnic:
> I fell off my motorbike and got some spare time lol

O_o glad it's nothing serious. OSS contributor - level hard ;)

PR #371, opened 2016-12-22
Update dry-types dependency to 0.9

timriley:
This should make it possible to work with rom-sql 0.9, which also depends on dry-types 0.9.


timriley:
As @flash-gordon summed up:

> actually not, ~> 0.8 is compatible with 0.9.X. ~> 0.8 is the same as >= 0.8, < 1.0. It's really easy to get confused because ~> 0.8.0 means the very different thing: >= 0.8.0, < 0.9

😬

PR #372, opened 2016-12-30
Replace ROM::Options with dry-initializer

flash-gordon:
@solnic here we're go :)
I've got an issue with dry-initializer under MRI 2.4, will see what Travis says

### DIFF
--- a/lib/rom/command.rb
+++ b/lib/rom/command.rb
@@ -1,6 +1,7 @@
 require 'dry/core/deprecations'
 require 'dry/core/class_attributes'
 
+require 'rom/types'
 require 'rom/initializer'
 require 'rom/pipeline'
 
@@ -35,8 +36,8 @@ module ROM
     # @attr_reader [Relation] relation The command's relation
     param :relation
 
-    CommandType = Dry::Types['strict.symbol'].enum(:create, :update, :delete)
-    Result = Dry::Types['strict.symbol'].enum(:one, :many)
+    CommandType = Types::Strict::Symbol.enum(:create, :update, :delete)
+    Result = Types::Strict::Symbol.enum(:one, :many)
 
     option :type, type: CommandType, optional: true
     option :source, reader: true, optional: true, default: -> c { c.relation }
--- a/lib/rom/command.rb
+++ b/lib/rom/command.rb
@@ -1,7 +1,7 @@
-require 'dry-initializer'
 require 'dry/core/deprecations'
 require 'dry/core/class_attributes'
 
+require 'rom/initializer'
 require 'rom/pipeline'
 
 require 'rom/commands/class_interface'
@@ -22,7 +22,7 @@ module ROM
   #
   # @private
   class Command
-    extend Dry::Initializer::Mixin
+    extend Initializer
     include Dry::Equalizer(:relation, :options)
     include Commands
     include Pipeline::Operator
@@ -137,15 +137,6 @@ module ROM
     # @api private
     def new(new_relation)
       self.class.build(new_relation, options.merge(source: relation))
-    end
-
-    # Instance options
-    #
-    # @return [Hash]
-    #
-    # @api public
-    def options
-      @__options__
     end
 
     private
--- a/lib/rom/command.rb
+++ b/lib/rom/command.rb
@@ -1,3 +1,4 @@
+require 'dry-initializer'
 require 'dry/core/deprecations'
 require 'dry/core/class_attributes'
 
@@ -21,6 +22,7 @@ module ROM
   #
   # @private
   class Command
+    extend Dry::Initializer::Mixin
     include Dry::Equalizer(:relation, :options)
     include Commands
     include Pipeline::Operator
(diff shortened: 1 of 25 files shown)


flash-gordon:
also we need to do something with https://github.com/dry-rb/dry-core/pull/7

lib/rom/command.rb
```diff
@@ -20,36 +20,31 @@ module ROM
...
+
+    CommandType = Dry::Types['strict.symbol'].enum(:create, :update, :delete)
+    Result = Dry::Types['strict.symbol'].enum(:one, :many)
+
+    option :type, type: CommandType, optional: true
+    option :source, reader: true, optional: true, default: -> c { c.relation }
+    option :result, reader: true, type: Result
     option :input, reader: true
-    option :curry_args, type: Array, reader: true, default: EMPTY_ARRAY
+    option :curry_args, reader: true, default: -> _ { EMPTY_ARRAY }
```
solnic: Only procs are supported? Seems like an overhead

lib/rom/command.rb
```diff
@@ -20,36 +20,31 @@ module ROM
...
-    # @attr_reader [Relation] relation The command's relation
-    attr_reader :relation
-
-    # @api private
-    def initialize(relation, options = EMPTY_HASH)
-      super
-      @relation = relation
-      @source = options[:source] || relation
-    end
-
```
solnic: Well, that's pretty cool :)

lib/rom/command.rb
flash-gordon: yes, only procs. I think, that's because of how the plugin system is implemented /cc @nepalez

lib/rom/command.rb
```diff
@@ -142,6 +137,15 @@ def new(new_relation)
...
 
+    # Instance options
+    #
+    # @return [Hash]
+    #
+    # @api public
+    def options
+      @__options__
+    end
+
```
solnic: Any chance we could somehow encapsulate this in a single place?

solnic:
> also we need to do something with dry-rb/dry-core#7

I took care of this. Released dry-core v0.2.2 with `ClassAttributes` and rom core master uses it already. So, `Option` is the very last bit...:)

solnic:
@flash-gordon I rebased this on top of master and resolved conflicts

lib/rom/command.rb
flash-gordon: yep, that's why this is WIP :P Just wanted to make it work. I think, I'll wrap initializer mixin with a module which includes instance methods, `#options` specifically

lib/rom/command.rb
nepalez: Yes, only procs are allowed by design
This was done by 2 reasons
The main one is that this way we prevent mutation of default values an instance was initialized with
The minor reason is that using procs, evaluated in new instance's scope allows setting defaults depending on other arguments

flash-gordon: we could wrap plain values with proc automatically. Mutation concerns can be addressed with `.frozen?` check, WDYT?

nepalez: Nope, freezing provides too strong limitation.

Actually what we do wanted is not avoiding mutation per se. It can be useful in many cases like one-off service objects. Freezing would break that cases totally. We wanted to prevent leeking mutations via defaults between various instances.

In february @marshall-lee convinced me to use procs by this simple example:

```ruby
require 'dry/initializer'

class A
  extend Dry::Initializer::Mixin

  param :name, default: proc { '...' }

  def call
    name << 'HAHA'
  end
end

3.times { puts A.new.call }

# => '...HAHA'
# => '...HAHA'
# => '...HAHA'
```

In case of no-proc instantiation (was implemented from the beginning, but removed now) this would provide something like:

```ruby
class B
  extend Dry::Initializer::Mixin

  param :name, default: '...'

  def call
    name << 'HAHA'
  end
end

3.times { puts B.new.call }

# => '...HAHA'
# => '...HAHAHAHA'
# => '...HAHAHAHAHAHA'
```

With a proc we do mutation after instantiation, but prevent leaking it outside (through a single object used as default value). Now all defaults are different objects, so various instances of A become independent from each other.

flash-gordon: I agree with this, but I meant something different:
```ruby
# error, use procs
param :foo, default: 'bar'

# works fine, doesn't raise an error if an object was frozen, 
# wraps it with a proc internally
param :foo, default: 'bar'.freeze

# works fine, booleans are frozen
param :foo, default: true
```

flash-gordon:
@solnic this is ready to go, I'll send PRs to rom-sql and rom-repository in the next year ;)

nepalez:
сб, 31 дек. 2016 г. в 14:24, Nikita Shilnikov <notifications@github.com>:

lib/rom/command.rb
nepalez: Ahh, I see. But your solution requires deep analysis of whether a value is frozen to the guts. Shallow freezing isn't enought. This in turn require more dependencies (more than zero which means infinitely more), or more code and more overhead. Tbh defaults are overhead by themselves, but we love them ;)

That's why I'm strongly against this kind of interface simplification

(Actually I've made a same try in May, not via analyzis of freezing, but value dup-ing under the hood. That turns out to became a disaster, and I've cut and burn it eventually)

solnic:
@flash-gordon great, I'm gonna check it out tomorrow (travelling today without my laptop)

lib/rom/command.rb
flash-gordon: I'd not go with a deep analysis, because no one stops you from doing like this
```ruby
cache = {}
param :cache, default: proc { cache }
```
From my point of view checking `.frozen?` is just enough to be sure that the user knows what he or she is doing. 
Just my thoughts, I don't insist on any changes :)

solnic:
I pushed [a cosmetic change](https://github.com/rom-rb/rom/pull/372/commits/8da75578c7c322f27bbbe1392874862e46aecb7c) before merging. Thanks @flash-gordon and @nepalez for helping with this, it's such a beautiful example of code reuse between dry/rom :)

lib/rom/setup/auto_registration_strategies/no_namespace.rb
```diff
@@ -1,13 +1,14 @@
...
 require 'dry/core/inflector'
+require 'rom/types'
 require 'rom/setup/auto_registration_strategies/base'
 
 module ROM
   module AutoRegistrationStrategies
     class NoNamespace < Base
-      option :directory, reader: true, type: Pathname
-      option :entity, reader: true, type: Symbol
+      option :directory, reader: true, type: PathnameType
```
nepalez: @solnic just for the clearance: you could use a "simplified" syntax as well (I've added it some time ago for compatibility to reform etc.)

```ruby
option :directory, PathnameType, reader: true
```

lib/rom/command_registry.rb
```diff
@@ -9,10 +10,7 @@ class CommandRegistry
     include Commands
 
     CommandNotFoundError = Class.new(KeyError)
-    RegistryType = Dry::Types::Definition
-                   .new(Registry)
-                   .constrained(type: Registry)
-                   .constructor { |r| r.is_a?(Registry) ? r : Registry.new(r, self.class.name) }
+    RegistryType = Types.Definition(Registry) | Types.Constructor(Registry) { |r| Registry.new(r, self.class.name) }
```
flash-gordon: Ha! :+1:

PR #374, opened 2017-01-05
Add before/after command hooks

solnic:
## Overview

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️  **BREAKING CHANGE** ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️

This makes it impossible to override `execute` to pre/post process input tuples, and makes `execute` a true implementation detail in adapters. This approach turned out to be problematic as we had to create additional command classes, configuring them was awkward, and it was inflexible. Instead of overidding `execute` this PR introduces a new concept of before/after hooks that are applied with correct arguments and in correct order.

This addition means two things:

- there's no longer any need to override `execute` (which was private anyway) in order to pre- or post-process input or result tuples
- hooks are based on class-level pre-configured options, or instance-level options that can be used at run-time
- initially we're going to support hooks as a list of command method names, and additional arguments can be provided for things like associations (explicit assoc name, or join keys etc.)

This feature is mostly for rom-sql, as its `associates` plugin has been difficult to use as it requires creating special command classes with overridden `execute`, which was an overkill; however, I'm pretty sure many other adapters can benefit from this.

### Hook types

- `before` is executed using curried args and call args, this allows us to prepare curried tuples based on additional args, the result of applying this hook is passed directly to `execute` method
- `after` is applied to the result of `execute` method, and whatever is returned becomes the actual command result, either a single tuple or an array with 1 or more tuples

### New API

``` ruby
# you can preconfigure a command class to always apply hooks
class CreateTask < ROM::Commands::Create[:memory]
  result :one
  relation :users
  before :associate_user

  def associate_user(task, user)
    task.merge(user_id: user[:id])
  end
end

# call with curried task and a user
create_task.with(title: "Task 1").call(id: 1, name: "Jane")
```

The actual cool part, and one of the most important reasons why this feature was created, is to be able to ask for a command with hooks dynamically at run-time. This will allow us to prepare all types of commands with additional tuple processing in places like repositories, without having to create special command *classes*:

``` ruby
# this is the same as the above, but at runtime
class CreateTask < ROM::Commands::Create[:memory]
  result :one
  relation :users

  def associate_user(task, user)
    task.merge(user_id: user[:id])
  end
end

# call with curried task and a user
create_task.before(:associate_user).with(title: "Task 1").call(id: 1, name: "Jane")
```

### TODO

- [x] add `before` and `after` command hooks
- [x] update graph to use hook-based command calls
- [x] add ability to provide additional args for hooks, ie `before(associate: :user)` where `:user` is an association name
- [x] add `Command#before` and `Command#after`
- [x] port `associates` rom-sql plugin to use this new API to make sure it does the job

### DIFF
--- a/lib/rom/command.rb
+++ b/lib/rom/command.rb
@@ -88,13 +88,47 @@ module ROM
     def call(*args, &block)
       prepared =
         if curried?
-          before_hooks.reduce(*curry_args) { |a, e| __send__(e, a, *args) }
+          before_hooks.reduce(*curry_args) do |a, e|
+            if e.is_a?(Hash)
+              hook_meth, hook_args = e.to_a.flatten
+              __send__(hook_meth, a, *args, **hook_args)
+            else
+              __send__(e, a, *args)
+            end
+          end
         else
-          before_hooks.reduce(*args) { |a, e| __send__(e, a) }
+          before_hooks.reduce(*args) do |a, e|
+            if e.is_a?(Hash)
+              hook_meth, hook_args = e.to_a.flatten
+              __send__(hook_meth, a, **hook_args)
+            else
+              __send__(e, a)
+            end
+          end
         end
 
       result = prepared ? execute(prepared, &block) : execute(&block)
-      finalized = after_hooks.reduce(result) { |a, e| __send__(e, a) }
+
+      finalized =
+        if curried?
+          after_hooks.reduce(result) do |a, e|
+            if e.is_a?(Hash)
+              hook_meth, hook_args = e.to_a.flatten
+              __send__(hook_meth, a, *args, **hook_args)
+            else
+              __send__(e, a, *args)
+            end
+          end
+        else
+          after_hooks.reduce(result) do |a, e|
+            if e.is_a?(Hash)
+              hook_meth, hook_args = e.to_a.flatten
+              __send__(hook_meth, a, **hook_args)
+            else
+              __send__(e, a)
+            end
+          end
+        end
 
       if one?
         finalized.first
--- a/lib/rom/command.rb
+++ b/lib/rom/command.rb
@@ -31,7 +31,7 @@ module ROM
     extend Dry::Core::ClassAttributes
     extend ClassInterface
 
-    defines :adapter, :relation, :result, :input, :register_as, :restrictable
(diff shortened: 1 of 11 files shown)


solnic:
@timriley @flash-gordon any thoughts on this one?

timriley:
From your description, this sounds really good. Having to override execute is user-unfriendly and inflexible. I like the ability to extend a command with hooks at runtime too.

solnic:
@timriley cool, FWIW I wanted to do this in rom 2.0.0 but ran out of time O_o but now it's a must-have because repos need this badly

lib/rom/command.rb
```diff
@@ -84,7 +86,25 @@ def execute(*)
...
+
+          result = prepared ? execute(prepared, &block) : execute(&block)
+
+          if curried?
+            apply_hooks(after_hooks, result, *args)
+          else
+            apply_hooks(after_hooks, result)
+          end
+        else
+          execute(*(curry_args + args), &block)
```
flash-gordon: so you do
```ruby
apply_hooks(before_hooks, *curry_args, *args)
```
and then
```ruby
execute(*(curry_args + args), &block)
```
I was curious which one is faster and found that `*(a + b)` is faster than `*a, *b` by ~20 percent in all the cases. :chart_with_upwards_trend:

solnic: yeah I know there are perf differences, I wanted to look into that closer but...you did that already, thank you :) we can tweak this in master now

lib/rom/command.rb
```diff
@@ -140,11 +170,33 @@ def new(new_relation)
...
     # @api private
     def composite_class
       Command::Composite
     end
+
+    # @api private
+    def apply_hooks(hooks, tuples, *args)
+      hooks.reduce(tuples) do |a, e|
+        if e.is_a?(Hash)
+          hook_meth, hook_args = e.to_a.flatten
```
flash-gordon: As we saw recently, `flatten(1)` is more predictable, it's also 20% faster! :grimacing:

PR #375, opened 2017-01-08
Add Gateway#transaction

flash-gordon:
Now we have a dummy transaction interface. It offers the only method `t.rollback!` which skips the rest of the current transaction, but does not roll back anything. In rom-sql calling `rollback!` will behave as expected


flash-gordon:
https://github.com/rom-rb/rom-sql/pull/128

PR #377, opened 2017-01-15
Add support for defining schema types for reading data

solnic:
This extends schema DSL with the ability to define "read" type that will
be used when a relation loads its tuples. By default we set this to a
noop passthrough-proc but if there are any attributes with a special
"read" type then we use relation's schema "relation hash" which is built
from schema attributes including "read" types.

This is a last-moment addition but it turned out we can't support
various custom DB types without this feature, and it was planned already
for rom 3.1.0 anyway.

Here's extended schema API:

``` ruby
class Users < ROM::Relation[:yaml]
  schema do
    # this means that the canonical db type is a string but we want an int
    # in resulting relation tuples
    attribute :id, Types::String, read: Types::Coercible::Int
    attribute :name, Types::String
  end
end
```

/cc @flash-gordon @AMHOL

### DIFF
--- a/lib/rom/plugins/command/schema.rb
+++ b/lib/rom/plugins/command/schema.rb
@@ -20,9 +20,9 @@ module ROM
 
               input_handler =
                 if default_input != Hash && relation.schema?
-                  -> tuple { relation.schema_hash[input[tuple]] }
+                  -> tuple { relation.input_schema[input[tuple]] }
                 elsif relation.schema?
-                  relation.schema_hash
+                  relation.input_schema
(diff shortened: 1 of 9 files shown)


solnic:
I should add that inferrer can be extended to infer `:read` types based on various criteria. Good example is jruby + sqlite which for some reason doesn't coerce date/time strings (as they are stored by sqlite) into date/time object, we can *easily* solve this kind of discrepancies between different rubies/dbs/drivers :) /cc @cflipse

PR #379, opened 2017-01-23
Allow multiple before/after hooks to be specified

cflipse:
A relatively common pattern is to allow a hook to be specified multiple times.  At present, calling a `before` or `after` multiple times will override the previous value.

I've attached a spec tweaked to demonstrate what I'd consider the less surprising behavior.  I'm not conversant enough to argue with the Dry::Intitializer macros to make this actually pass, but I wanted to get this noted before we go live with the hooks.


solnic:
I finalized this in master :)

PR #386, opened 2017-03-05
Update dry-initializer to 1.3

flash-gordon:
One issue related to schema caching showed up after that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,3 +52,5 @@ group :tools do
   gem 'mutant'
   gem 'mutant-rspec'
 end
+
+gem 'dry-initializer', git: 'https://github.com/dry-rb/dry-initializer.git', branch: 'v1.3.0'
--- a/Gemfile
+++ b/Gemfile
@@ -52,5 +52,3 @@ group :tools do
   gem 'mutant'
   gem 'mutant-rspec'
 end
-
-gem 'dry-initializer', git: 'https://github.com/dry-rb/dry-initializer.git', branch: 'v1.3.0'
--- a/lib/rom/command.rb
+++ b/lib/rom/command.rb
@@ -204,7 +204,7 @@ module ROM
 
     # @!attribute [r] source
     #   @return [Relation] The source relation
-    option :source, reader: true, optional: true, default: -> c { c.relation }
+    option :source, reader: true, optional: true, default: -> { relation }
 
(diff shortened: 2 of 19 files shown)


solnic:
How's performance?

solnic:
re schema caching, we need to do something smarter because `input_schema` needs to be recalculated when `schema` was projected in some way

flash-gordon:
re performance didn't test, was going to ask you haha. I mean what should I benchmark?

solnic:
Tests run for me almost 2 x slower

flash-gordon:
ref https://github.com/dry-rb/dry-initializer/issues/25

solnic:
I think our ultimate solution for dry-initializer + rom-rb should be disabling `undefined` functionality, it adds complexity that isn't justified in any way. WDYT?

flash-gordon:
@solnic it's kinda hardwired inside the gem atm, we could have our own getters though, I'll give it a try

flash-gordon:
@solnic dry-initializer now has the `undefined: false` option that we use to disable `UNDEFINED` in instance variables,

```ruby
2.4.0 :006 > require 'dry-initializer'
 => true
2.4.0 :007 > Dry::Initializer
 => Dry::Initializer
2.4.0 :008 > class Foo
2.4.0 :009?>   extend Dry::Initializer[undefined: false]
2.4.0 :010?>   option :foo, optional: true
2.4.0 :011?>   end
 => #<Module:0x007ff24e116b90>
2.4.0 :012 > Foo.new
 => #<Foo:0x007ff24e105d90 @__options__={}, @foo=nil>
```

Resolved default values stored in `@__options__`:
```ruby
2.4.0 :002 > class Bar
2.4.0 :003?>
2.4.0 :004 >     extend Dry::Initializer[undefined: false]
2.4.0 :005?>   option :bar, default: -> { 1 }
2.4.0 :006?>   end
 => #<Module:0x007fb0c50145a0>
2.4.0 :007 > Bar.new
 => #<Bar:0x007fb0c4835690 @__options__={:bar=>1}, @bar=1>
```

I'm merging this, kudos to @nepalez for helping with this

P.S. Specs are still slower than with 0.11, but that's because of class-level code evaluated on subclassing `Relation` and/or `Command`, this won't affect the run-time performance.

Issue #349, opened 2016-08-05
Extract mapper-related features into rom-mapper 


wmaciejak:
Issue discovered while upgrading ROM to version 2.0.0. 

```
rom (= 2.0.0) was resolved to 2.0.0, which depends on
      rom-mapper (~> 0.4.0)
```

Other, related issue - when you will try use mapper on combined records, then you will see warnings like this

```
[1] pry(#<UserRepository>)> users.combine(one: { settings: [settings.for_org(org), id: :user_id] }).as(:entity).to_a
/Users/user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/transproc-0.4.0/lib/transproc/hash.rb:246: warning: wrong element type Hash at 0 (expected array)
/Users/user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/transproc-0.4.0/lib/transproc/hash.rb:246: warning: ignoring wrong elements is deprecated, remove them explicitly
/Users/user/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/transproc-0.4.0/lib/transproc/hash.rb:246: warning: this causes ArgumentError in the next release
=> [{:settings=>{:active=>false}}, {:settings=>{:active=>false}}]
```

Mapper are very simple

```
module UserMappers
  class Entity < ROM::Mapper
    relation :users
    register_as :entity
    reject_keys true

    attribute :id
    attribute :code
    attribute :name
    attribute :settings do |settings|
      settings ? settings[:value].deep_symbolize_keys : { active: false }
    end
  end
end
```

solnic:
Done in master

PR #387, opened 2017-03-09
[WIP] Move mapper logic into rom-mapper

GustavoCaso:
#349 
@solnic 
As we talked about this an early WIP for extracting the logic of `Mappers` to `rom-mapper` and register via plugin, https://github.com/rom-rb/rom-mapper/pull/42.

I haven't actually tested myself but I wanted to have a place where we could discuss further actions.

Following how the plugin system work I hope I have it understand it, I remove all the code related to Mappers to is PluginConfiguration.

This is PR in `rom-mapper` adding the behavior for register mapper


solnic:
So, the PR on rom-mapper side looks good, so last thing to do is to update things here so the build can pass :)

PR #389, opened 2017-04-02
Add schema plugins

flash-gordon:


flash-gordon:
I've found this would be a really nice addition, this will reduce the boilerplate on defining schemas significantly and will allow to add some cool stuff to `rom-sql`. Specifically I want to have auto-index attributes in schemas because you normally want to have all the FKs to be indexed.

solnic:
This is cool indeed. I wanted to add it too. So +1

flash-gordon:
@solnic glad to see you :) I'm gonna make it work with relation finalizers today/tomorrow so that it will be possible to load plugins globally, then this will be ready for going in master. Gotta say the plugin infrastructure is a bit messy and I guess will require some refactoring before 4.0.

solnic:
> Gotta say the plugin infrastructure is a bit messy and I guess will require some refactoring before 4.0.

Yeah it should be improved in 4.0

solnic:
@flash-gordon schema needs to be established already because it is needed in `view` DSL, so you can't postpone it until finalization :/

flash-gordon:
we infer schemas at the finalization step thus `viewDSL` cannot depend on it without having corner cases where `view` syntax breaks for unclear reasons

solnic:
@flash-gordon hmm, right, view schemas are [finalized here](https://github.com/rom-rb/rom/blob/master/lib/rom/relation/class_interface.rb#L340-L346). I guess if you move things around so that schema dsl can be extended in the right moment, things will continue to work.

flash-gordon:
> we infer schemas at the finalization step thus viewDSL cannot depend on it without having corner cases where view syntax breaks for unclear reasons

I actually saw those errors, now just things clicked in place :)

flash-gordon:
@solnic phew! I store an unresolved schema in `@schema_proc` of a relation class. Overall, postponing schema initialization wasn't hard, in `Relation.view` I've changed `schema.project(*args[1])` to ` -> _ { schema.project(*args[1]) }`

There is a bunch of related changes in rom-sql and I'll push them later today. While I was working on this I realized how hardly we need some sort of bootstrapping pipeline with events, callbacks and things like that. It's very tough to learn how every single hook in rom works because they works differently. 

I wish it would be 
```ruby
ROM::Bootstrap.before(:schema_load, relation: :users) do |container, schema_dsl: , **other_stuff|
  schema_dsl.extend(Timestamps::SchemaDSL)
end
```
^just a sketch, but wdyt?

solnic:
@flash-gordon sounds good!

re bootstraping, it crossed my mind many times, but it's such a complex part of rom that I never had the energy to tackle it. There was also a long learning curve to understand what's involved in this process, and now we're in a much better position to come up with a nicer approach. We can try to do it in 4.0.

lib/rom/plugin_base.rb
```diff
@@ -28,6 +28,11 @@ def relation?
       type == :relation
     end
 
+    # @api pribate
```
artofhuman: typo `private`

flash-gordon: >.< thanks!

core/lib/rom/plugins/schema/timestamps.rb
```diff
@@ -0,0 +1,58 @@
...
+
+        # @api private
+        def self.apply(schema, options)
+          type = options.fetch(:type, Types::Time)
+          names = options.fetch(:attributes, DEFAULT_TIMESTAMPS)
+          attributes = names.map { |name| type.meta(name: name, source: schema.name) }
+
+          schema.attributes.concat(
+            schema.class.attributes(attributes, schema.attr_class)
+          )
```
solnic: Is `concat` safe here?

core/lib/rom/relation/class_interface.rb
```diff
@@ -94,23 +94,35 @@ def schema(dataset = nil, as: nil, infer: false, &block)
...
-            name,
+            relation_name,
             schema_class: schema_class, attr_class: schema_attr_class, inferrer: inferrer,
             &block
           )
 
-          @schema = dsl.call
+          @schema_proc = dsl.method(:call).to_proc
+        elsif dataset.is_a?(Schema)
+          @schema = dataset
```
solnic: How come `dataset` can be a schema instance? Looks like a workaround of some problem?

core/lib/rom/plugins/schema/timestamps.rb
flash-gordon: yeah, this is what we need, add attributes to an existing schema before it gets frozen but without using the dsl

core/lib/rom/setup/finalize/finalize_relations.rb
```diff
@@ -57,9 +57,19 @@ def build_relation(klass, registry)
...
         #       where klass' gateway points to non-existant repo
         gateway = @gateways.fetch(klass.gateway)
+
+        if klass.schema_proc && !klass.schema
+          plugins = schema_plugins
+          resolved_schema = klass.schema_proc.call do
+            plugins.each { |plugin| app_plugin(plugin) }
+          end
+
+          klass.schema(resolved_schema)
```
solnic: ...ahh it's here...can we add `set_schema!(schema)` instead? This could be used in `inherited` hook to where we inherit schema from a parent class.

solnic:
Well, this is pretty awesome. I've left a couple of comments (small things to address).

flash-gordon:
@solnic done. I actually removed setting a schema in the `inherited` hook because it's not defined when the hook gets called.

solnic:
Looks like rom 4.0 will be more awesome than we thought ;)

PR #391, opened 2017-05-05
Expand AutoRegistrationStrategies::CustomNamespace

Krule:
Hello ROM people.

Some time ago I wanted to autoload some namespaced `ROM::Relation` I have had defined in my project and while everything looked fine and according to namespace path, I was surprised when this did not quite work out.

After some searching and asking on the gitter, I decided to take a look at the code (duh) and figured out autoloader takes only `File.basename(file, ".rb")` as constant name. So, in order to have everything working properly, I could not use `autoload(namespace: "Persistance")` with:

```ruby
# ./lib/persistence/relation/user.rb
module Persistance
  module Relation
    module Users < ROM::Relation[:sql]
      # stuff...
    end
  end
end
```

... as this would attempt to load `Persistence::Users` instead of `Persistence::Relation::Users` and that, of course, would fail. Now, after I figured this out, I modified my code so it would work as expected.

The end!

Well, maybe. Since this was a bit confusing for me, I imagine it might as well be for other people, so I would like to offer a PR which might remedy the situation.

### DIFF
--- a/lib/rom/setup/auto_registration.rb
+++ b/lib/rom/setup/auto_registration.rb
@@ -29,7 +29,7 @@ module ROM
     option :globs, default: -> {
       Hash[
         component_dirs.map { |component, path|
-          [component, directory.join("#{ path }/**/*.rb")]
+          [component, directory.join("#{path}/**/*.rb")]
         }
       ]
     }
--- a/lib/rom/setup/auto_registration.rb
+++ b/lib/rom/setup/auto_registration.rb
@@ -66,6 +66,7 @@ module ROM
               file: file, directory: directory, entity: component_dirs.fetch(entity)
             ).call
           end
+
         Dry::Core::Inflector.constantize(klass_name)
       end
     end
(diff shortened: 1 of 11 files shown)


solnic:
Thanks for looking into this and opening PR. It's hard to make everybody happy, but that's why we have strategies to cover different use cases.

Anyhow, re PR - will this continue to work like before, as in when you don't use component names inside namespaces, so ie `My::Stuff::Users` rather than `My::Stuff::Relations::Users`? I see you modified existing spec, so I'm not sure if it's a breaking change or not.

Krule:
The modification in specs was adding new use cases, leaving old ones unchanged (other than putting them in `context`). So, old behaviour should remain unchanged provided constant exists.

Looking at code now, I realise I have not provided anything happening if fragments are exhausted without actually finding the constant (when it's something else). I think the current implementation would throw `NameErorr`. This I would like to replace with a more descriptive error.

I will make some more modifications soon and then ask for review 😄

Krule:
Updated the code, limiting the search to directory and providing more helpful error message.

Krule:
Ah yes, to clarify. This is should not be a breaking change.

solnic:
Thanks man, I'll check it out on Friday

solnic:
It's merged. Thanks again!

PR #398, opened 2017-05-28
Remove relation proxy and port its features into core APIs

solnic:
Holy 💩 I've made it 🎉 

This is actually "just" a first step, it's huge but it's a good start. Once this is in master I'll focus on cleaning things up. There are already a bunch of cool improvements in this PR, but it'll be even better eventually.


core/lib/rom/relation.rb
```diff
@@ -81,6 +84,30 @@ class Relation
...
     }
 
+    # @!attribute [r] mappers
+    #   @return [MapperRegistry] an optional mapper registry (empty by default)
+    option :mappers, default: -> { MapperRegistry.new }
+
+    # @!attribute [r] auto_struct
+    #   @return [TrueClass,FalseClass] Whether or not tuples should be auto-mapped to structs
+    #   @api private
+    option :auto_struct, reader: true, default: -> { false }
```
flash-gordon: reader is `:public` by default
https://github.com/dry-rb/dry-initializer/blob/0d2fa01b11a0669c61eddd0fbc4e18d7aa4094a8/lib/dry/initializer/attribute.rb#L48

solnic: Ahh good point. TBH I didn't pay much attention to method's visibility but I'll address this eventually.

PR #401, opened 2017-05-29
Remove mapper logic from core

GustavoCaso:
@solnic Hope this attempt is the one.

Move logic for creating mappers from `rom-core` to `mappers`, by using the plugin system


solnic:
This is cool. I'll merge this in but there will be another small task related to this. I'll report it shortly.

PR #406, opened 2017-06-09
Dynamic command registry

solnic:
Just like with mapper registry, this turns command registry into a dynamic registry that can compile commands on the fly and encapsulate access to commands in `Relation#command`. All relations have access to their command registries as a consequence.


core/lib/rom/command_compiler.rb
```diff
@@ -0,0 +1,260 @@
...
+    #   @param adapter [Symbol] The adapter identifier
+    #   @param ast [Array] The AST representation of a relation
+    #   @param plugins [Array<Symbol>] A list of optional command plugins that should be used
+    #   @param meta [Hash] Meta data for a command
+    #
+    #   @return [Command, CommandProxy]
+    #
+    # @api private
+    def [](*args)
+      fetch_or_store(args.hash) do
```
flash-gordon: @solnic didn't we lose `freeze` somewhere?

solnic: Freezing what exactly?

flash-gordon: the command class :)

solnic: Oh I didn't know we freeze command classes. IIRC I *planned* to add freezing them (same with mappers and relations).

flash-gordon: remember this https://github.com/hanami/model/pull/405

flash-gordon: @solnic I just wanted to know if you still remember it :)

solnic: Ah OK :) yeah I do remember, I'll do it once I'm done with local caches refactor in #408

PR #408, opened 2017-06-10
Remove global caches

solnic:
This refactors/unifies registries further and removes global caches from mapper-related compilers/builders.

### TODO

- [x] unify registries wrt handling options and their constructors
- [x] introduce `ROM::Cache`
- [x] add common `cache` option to registries
- [x] pass caches to mapper and command registries when setting up a container
- [x] refactor mapper compiler and its collaborators so that they use local caches


solnic:
@flash-gordon still WIP but please take a look if you can and let me know if this makes sense

core/lib/rom/relation_registry.rb
```diff
@@ -2,17 +2,9 @@
 
 module ROM
   class RelationRegistry < Registry
-    def initialize(elements = {})
+    def initialize(elements = {}, options = {})
       super
       yield(self, elements) if block_given?
     end
```
solnic: I'm gonna remove this custom constructor as we can now use `Registry#map` to achieve the same thing (it's needed in `FinalizeRelations`

core/lib/rom/registry.rb
```diff
@@ -12,12 +12,39 @@ class Registry
     include Enumerable
     include Dry::Equalizer(:elements)
 
-    param :elements, default: -> { Hash.new }
+    param :elements
+
+    option :cache, reader: true, default: -> { Cache.new }
+
+    def self.new(*args)
```
flash-gordon: I wonder if we really need this, dry-initializer should work fine with inheritance, so perhaps we might remove `param :elements` from the descendants instead. Not sure though

core/lib/rom/container.rb
```diff
@@ -115,11 +116,16 @@ class Container
     #   @return [Hash] A hash with configured custom mappers
     attr_reader :mappers
 
+    # @!attribute [r] caches
+    #   @return [Hash] A hash with configured caches for rom components
+    attr_reader :caches
+
     # @api private
     def initialize(gateways, relations, mappers, commands)
+      @caches = { mappers: Cache.new, commands: Cache.new }.freeze
```
flash-gordon: 👍

core/lib/rom/registry.rb
solnic: @flash-gordon it did not work, hence this custom `.new`

flash-gordon: yeah, I guess there's _a reason_ 😄  I'll have a look later

solnic: the problematic part is that both `elements` and `options` are hashes, so when you don't pass anything, then default elements and options become a single hash, which ends up with an empty registry with `options[:elements]`. Maybe this helps.

flash-gordon: oh wow

Issue #402, opened 2017-05-30
Extract registering arbitrary mappers from rom/mapper to rom/core


solnic:
So in #401 we moved everything related to mappers to rom/mapper, but there's one caveat there - we have a way of registering arbitrary mapper objects via `Mapper::ConfigurationDSL#register`. So the trick is that the code that is specific to rom/mapper is only `mappers { define(...) { ... } }` which exposes mapper DSL, but `mappers { register(..) { ... } }` is just a way of registering anything as a mapper, and this should be part of core.

GustavoCaso:
I can help with this one

solnic:
Closing this as it isn't easily solvable at this point. I'll revisit it after 4.0.

PR #409, opened 2017-06-10
[WIP] Extract behaviour for registering arbitrary mappers to rom-core

GustavoCaso:
This try to solves #402 

This is a first iteration over the problem itself but wanted to share so people can comment on it.


GustavoCaso:
@solnic maybe we can discuss this one when you have some time.

solnic:
Sorry but I'm going to close this. It seems like it's not worth the effort and creates too much indirection. We can revisit this later in 4.x or 5.0 and this PR can serve as a reference.

PR #422, opened 2017-06-20
Improve association lookup

solnic:


core/lib/rom/associations/definitions/abstract.rb
```diff
@@ -51,15 +56,45 @@ class Abstract
...
+        def self.process_options(target, options)
+          through = options[:through]
+
+          if through
+            options.update(through: ThroughIdentifier[through, target.relation, options[:assoc]])
+          end
+
+          options.update(name: target.relation)
+
+          options
```
flash-gordon: technically, you don't need this line, `updates` returns the updated hash anyway, otherwise, I'd prefer to go with `options[:name] = target.relation` because this won't allocate an intermediate hash instance '{ name: target.relation }'

solnic: Thanks, I've changed it based on your feedback.

Issue #425, opened 2017-06-21
Mapper::Builder doesn't set `register_as`


solnic:
This was originally reported here https://github.com/rom-rb/rom-repository/issues/83

...and it boils down to adding `klass.register_as(name)` [right here](https://github.com/rom-rb/rom-repository/issues/83).

PR #429, opened 2017-06-26
Added register_as when defining a mapper

GustavoCaso:
Solves #425

### DIFF
--- a/mapper/lib/rom/mapper/builder.rb
+++ b/mapper/lib/rom/mapper/builder.rb
@@ -25,6 +25,7 @@ module ROM
           end
 
         Dry::Core::ClassBuilder.new(name: class_name, parent: parent_class).call do |klass|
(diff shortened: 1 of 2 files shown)


solnic:
We need a spec for this too :)

GustavoCaso:
@solnic after adding the specs I realize something, correct if I'm wrong what happens if a user instead of registering a mapper with the mapper DSL instead he creates a custom Mapper class and forget to add the `register_as` since we no control that registration process we maybe need to raise an error indicating to the user that they are missing the `register_as`. What do you think?

solnic:
@GustavoCaso IIRC we default to relation identifier when there's no `register_as` defined on a mapper class, which means we'll get an error saying that same mapper was registered more than once in a situation when there's more than one mapper class without `register_as`, so I'd say let's leave it as it is now.

PR #431, opened 2017-06-27
Auto-generate #with

flash-gordon:
It's a kind of PoC, @solnic do you think it's a good idea to have `with` by default?

P.S. We cannot have it in dry-initializer because it has an API for renaming option keys

### DIFF
--- a/core/lib/rom/command_compiler.rb
+++ b/core/lib/rom/command_compiler.rb
@@ -104,11 +104,6 @@ module ROM
     alias_method :[], :call
 
     # @api private
-    def with(new_options)
-      self.class.new(gateways, relations, commands, notifications, options.merge(new_options))
-    end
-
-    # @api private
     def type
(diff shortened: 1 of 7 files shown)


solnic:
Yes it's a good idea, I already had cases where I defined it yet-another-time but with a bug and things got broken, so it'd be good to have it solved once and for all.

flash-gordon:
What should I do with `Schema#with`? It's marked as a public API but don't think anyone is going to miss it

solnic:
> What should I do with Schema#with? It's marked as a public API but don't think anyone is going to miss it

It's fine if it's gone, just make sure it's deprecated in release-3.0 branch

flash-gordon:
@solnic I meant the doc is gone, the method is still there, now auto-generated 😬

solnic:
@flash-gordon I believe we're masters of using yard features nobody knows about, so we can still document it :)

Issue #434, opened 2017-07-02
Raise a meaningful error if relation name is one of the reserved names


solnic:
For now we know that `relations` can't be used as a name of a relation. Would be good to look into code and figure out if there are any other names that could break things.

GustavoCaso:
@solnic I'm happy to help with this one

solnic:
@GustavoCaso cool. Let's start with `relations` then. We can then add more reserved names easily whenever we discover them.

PR #437, opened 2017-07-05
Raise exception for invalid relation name

GustavoCaso:
Solves #434 

@solnic after playing with the code I found this to be a good place for raising the error 😄 

Although the test infers the name from the `schema` macro I would have liked to test creating a `Relations` class but not under the namespace of `Test` but the `ConstantLiker` wouldn't allow me.

Hope this helps.

### DIFF
--- a/core/lib/rom/relation/class_interface.rb
+++ b/core/lib/rom/relation/class_interface.rb
@@ -96,9 +96,7 @@ module ROM
           ds_name = dataset || schema_opts.fetch(:dataset, default_name.dataset)
           relation = as || schema_opts.fetch(:relation, ds_name)
 
-          if INVALID_RELATIONS_NAMES.include?(relation.to_sym)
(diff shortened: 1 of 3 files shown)


core/lib/rom/relation/class_interface.rb
```diff
@@ -277,6 +284,12 @@ def default_schema(klass = self)
       def name
         super || superclass.name
       end
+
+      private
+
+      def invalid_relation_msg(relation)
+        "Relation name: #{relation} is a protected word, please use another relation name"
+      end
```
solnic: This could be encapsulated in the exception class itself

GustavoCaso: Totally right 😄

GustavoCaso:
@solnic sorry for the delay. Hope this is better now 😄

Issue #435, opened 2017-07-02
Date based restriction doesn't work with command macro


solnic:
_From @daniels on June 19, 2017 12:27_

When trying to use a time/date based restriction from a commands macro, the DateTime or Time object passed in as a parameter is converted to an Array and then splatted against the filter method, causing an ArgumentError.

The SQL-relation `:requests` has a `:created_at` column, and `Requests#expired` is defined:
```ruby
def expired(expiration_time=default_expiration_time)
  where{ created_at < expiration_time }
end
```

The commands macro is used like this:
```ruby
r = Class.new(ROM::Repository[:requests])  do
  commands delete: [:by_pk, :expired]
end.new(rom) 
```

And calling the generated method:
```ruby
r.delete_expired(Time.now)
#=> ArgumentError: wrong number of arguments (given 10, expected 0..1)
```

_Copied from original issue: rom-rb/rom-repository#93_

solnic:
Thanks for reporting this. I'll fix it in 1.x and port it to rom-rb/rom master.

GustavoCaso:
I would like to help with this one

solnic:
@GustavoCaso it's all yours (:

PR #440, opened 2017-08-01
Avoid splat arguments if they are coercible to array in repository commands macro

GustavoCaso:
Solves #435 

To avoid splatting the arguments if they are coercible into Array.

### DIFF
--- a/repository/lib/rom/repository/class_interface.rb
+++ b/repository/lib/rom/repository/class_interface.rb
@@ -113,7 +113,13 @@ module ROM
             if changeset.respond_to?(:commit)
               changeset.commit
             else
-              root.command(type, **opts).public_send(view_name, *view_args).call(*input)
+              #   in case of view_args been coercible to an array we do not want to splat it
+              #   example: Date and Time objects
+              if view_args.respond_to?(:to_a)
+                root.command(type, **opts).public_send(view_name, view_args).call(*input)
+              else
+                root.command(type, **opts).public_send(view_name, *view_args).call(*input)
+              end
             end
           end
         end
--- a/repository/lib/rom/repository/class_interface.rb
+++ b/repository/lib/rom/repository/class_interface.rb
@@ -113,13 +113,8 @@ module ROM
             if changeset.respond_to?(:commit)
               changeset.commit
             else
-              #   in case of view_args been coercible to an array we do not want to splat it
-              #   example: Date and Time objects
-              if view_args.respond_to?(:to_a)
-                root.command(type, **opts).public_send(view_name, view_args).call(*input)
-              else
-                root.command(type, **opts).public_send(view_name, *view_args).call(*input)
-              end
+              view_args = [view_args] unless view_args.respond_to?(:to_ary)
+              root.command(type, **opts).public_send(view_name, *view_args).call(*input)
             end
           end
         end
--- a/repository/spec/integration/command_macros_spec.rb
+++ b/repository/spec/integration/command_macros_spec.rb
@@ -191,5 +191,17 @@ RSpec.describe ROM::Repository, '.command' do
       deleted_name = repo.delete(1)
       expect(deleted_name).to eql('Jane Doe')
     end
+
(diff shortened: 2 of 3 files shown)


repository/lib/rom/repository/class_interface.rb
```diff
@@ -113,7 +113,13 @@ def define_restricted_command_method(type, views, **opts)
             if changeset.respond_to?(:commit)
               changeset.commit
             else
-              root.command(type, **opts).public_send(view_name, *view_args).call(*input)
+              #   in case of view_args been coercible to an array we do not want to splat it
+              #   example: Date and Time objects
+              if view_args.respond_to?(:to_a)
+                root.command(type, **opts).public_send(view_name, view_args).call(*input)
```
solnic: We must splat if this view's arity is > 1. You can check it via `method(view_name).arity`.

GustavoCaso: After some investigation and trial error, relaying in the `arity` is not going to be that simple.

Trying to access the arity value from `method(view_name).arity` since `view` is `AutoCurry` the arity is always `-1` .
Playing a little bit more I found a way to access the `arity` through out `Rom::Relation::Curried`

How I get the Curried Relation by executing this line:
`root.command(type, **opts).public_send(view_name)`

I get a Curried Relation, which happens to have an arity public API.

But I notice a strange behavior if the custom method from the relation have default values like this:

```ruby
def expired(expiration_time=Time.now)
  where{ created_at < expiration_time }
end
```
Instead of returning a  `Rom::Relation::Curried` I get a `ROM::Commands::Composite` which do not have a public api for `arity`

@solnic is there something I'm missing?

Also just for fun I played a bit with arity from ruby
```ruby
class C
 def one;    end
 def two(a); end
 def three(a=1); end
 def four(*a);  end
 def five(a, b); end
 def six(a, b, *c);    end
 def seven(a, b, *c, &d); end
end

c = C.new

c.method(:one).arity
=> 0
c.method(:two).arity
=> 1
c.method(:three).arity
=> -1
c.method(:four).arity
=> -1
c.method(:five).arity
=> 2
c.method(:six).arity
=> -3
c.method(:seven).arity
=> -3
```

daniels: What is the API for calling the command if the view has arity > 1? (There aren't any specs for such calls as far as I can see.)

Should the view args be wrapped in an array like so?:

```ruby
repo.delete_by_arity2_restriction([view_arg1, view_arg2], input)
```

(If it's not wrapped, I guess it would be splatted into `input` rather than the view_arg at the beginning of the defined command method?)

If it should be wrapped, could you maybe just wrap any non-array object in an array before splatting? Checking either by `#is_a?(Array)` or `#responds_to?(:to_ary)`:

```ruby
view_args = [view_args] unless view_args.responds_to?(:to_ary)
root.command(type, **opts).public_send(view_name, *view_args).call(*input)
```

Or even just by doing a splat reassignment:

```ruby
*view_args = view_args
root.command(type, **opts).public_send(view_name, *view_args).call(*input)
```

(Splat assignment only tries implicit conversion (`#to_ary`) and doesn't call `#to_a`, in contrast to the splatting of the argument on the next line which calls `#to_a` on view_args if `#to_ary` is not defined. But maybe this is too obscure to make the reason behind the code above clear.)

daniels: To avoid the reassignment you could even do the splatting in the original assignment like so:

```ruby
define_method(meth_name) do |*args|
  (*view_args), *input = args
  # ...
end
```

solnic: I'll wrap this up folks, it's pretty clear this is a bit more challenging to fix.

GustavoCaso: Thanks @solnic I'm intrigued actually

solnic:
Closing this in favor of https://github.com/rom-rb/rom/pull/450

PR #441, opened 2017-08-01
Generic JSON read/write type

Krule:
Hi, following suggestion on Gitter, https://gitter.im/rom-rb/chat?at=597fb669210ac269204184f4
I would like to propose a generic json read/write type: `Coercible::Json` together with:

- `Coercible::Json()`
- `Coercible::JsonHash()`
- `Coercible::HashJson()`

### DIFF
--- a/core/lib/rom/types.rb
+++ b/core/lib/rom/types.rb
@@ -47,7 +47,7 @@ module ROM
       end
 
       def JSON(symbol_keys: false)
-        self.JsonWrite.meta(read: self.JsonRead(symbol_keys))
+        self.HashJson.meta(read: self.JsonHash(symbol_keys))
       end
     end
   end
--- a/core/lib/rom/types.rb
+++ b/core/lib/rom/types.rb
@@ -27,7 +27,7 @@ module ROM
     end
 
     module CoercibleMethods
-      def JsonRead(symbol_keys = false, type = Types::Hash)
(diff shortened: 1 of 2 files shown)


Krule:
Also, if there is something wrong with the approach I have used, pointers would be appreciated. 😄

core/lib/rom/types.rb
```diff
@@ -22,5 +25,30 @@ def ForeignKey(relation, type = Types::Int)
         type.meta(foreign_key: true, target: relation)
       end
     end
+
+    module CoercibleMethods
+      def JsonRead(symbol_keys = false, type = Types::Hash)
+        Types.Constructor(type) do |value|
+          next Hash[value] if value.respond_to?(:to_hash)
```
flash-gordon: I don't think we need this check in ROM 4.0, `value` is expected to string and we don't process it twice. I hope 😬 However, in 3.0 it's required for rom-repo

Krule: @flash-gordon So, maybe a different PR for 4.0 and getting this one into 3.X?

Krule: Here it is: https://github.com/rom-rb/rom/pull/446

flash-gordon:
Thanks!

solnic:
Please rename `Json` to `JSON`

PR #445, opened 2017-08-15
Fix Relation#combine when using multi-key hashes

flash-gordon:
@solnic mind to test it? Works smooth on my side


solnic:
LGTM 👍

PR #446, opened 2017-08-15
Generic JSON read/write type for release-3.0

Krule:
Implement https://github.com/rom-rb/rom/pull/441 for `release-3.0`.

PR #447, opened 2017-08-17
Types/coercible json naming change

Krule:
**Addressing**: https://github.com/rom-rb/rom/pull/446#discussion_r133506000

Rename CoercibleMethods `HashJson -> HashJSON`, `JsonHash -> JSONHash`

Refactor attributes to keyword arguments.
- Add specs for usage of both individually.
- Add spec to cover case when Input JSON is invalid for `JSONHash`.

PR for 4.X will follow shortly.

@solnic I am also thinking about extending this a bit more by making it possible for JSON serializer to be plugged in. Any thoughts on idea? cc: @flash-gordon


solnic:
Thanks man :)

PR #450, opened 2017-08-19
Fix splatting args in command macros

### LINKED ISSUE #435: Date based restriction doesn't work with command macro

solnic:
Refs #440 
Closes #435

### DIFF
--- a/repository/lib/rom/repository/class_interface.rb
+++ b/repository/lib/rom/repository/class_interface.rb
@@ -113,7 +113,13 @@ module ROM
             if changeset.respond_to?(:commit)
               changeset.commit
             else
-              root.command(type, **opts).public_send(view_name, *view_args).call(*input)
+              #   in case of view_args been coercible to an array we do not want to splat it
+              #   example: Date and Time objects
+              if view_args.respond_to?(:to_a)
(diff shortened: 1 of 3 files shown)


solnic:
I merged this in already because it does fix the original #435 issue. I will refactor and improve command macros so that specific command types are handled in a better way. This should fix problems with relation views with `-1` arity too.

PR #451, opened 2017-08-23
Use dry-initializer v2.0

flash-gordon:


core/lib/rom/initializer.rb
```diff
@@ -57,8 +56,10 @@ module InstanceMethods
       #
       # @api public
       def options
-        @__options__
+        self.class.dry_initializer.attributes(self)
```
solnic: why does it require passing `self` in? I'd expect this type of info to be memoized at instance level, otherwise it's an overkill to process this every time `options` method is called.

flash-gordon:
@solnic this can go to master

PR #452, opened 2017-08-24
Add tests demonstrating issue with overriding struct attribute methods

timriley:
In working through a real app with the rom 4 betas, I noticed an issue with the way auto-generated structs in struct namespaces work.

There are 2 (related) aspects to this:

1. It's not possible to define overriding attribute methods on structs and refer to the original attribute value
2. Which also means when a struct class inherits from _another_ struct class, the superclass' overriding attribute method is not accessible

I _think_ the problem starts with these lines in `struct_compiler.rb`:

```ruby
          build_class(name, ROM::Struct, ns) do |klass|
            attributes.each do |(attr_name, type)|
              klass.attribute(attr_name, type)
            end
          end
```

This is going to define dry-struct attributes on the class that it builds.

Then if we go over to dry-struct's source and look at at this part of `class_interface.rb`:

```ruby
      def attributes(new_schema)
        check_schema_duplication(new_schema)

        schema schema.merge(new_schema)
        input Types['coercible.hash'].public_send(constructor_type, schema)

        attr_reader(*new_schema.keys)
```

When attributes are defined on a dry-struct, an `attr_reader` is _always_ created for each attribute.

The problem in this case here in rom is that the classes that are defined for auto-structs in a namespace all _inherit_ from the user's own custom struct classes. Since these `attr_readers` are defined in the subclass, the overridden attribute methods in the parent class are no longer reachable.

That's the class inheritance side of the problem (point 2 in my list above).

The other side of the problem is about accessing the original value when defining this overriding attribute methods in the first place. It makes a lot of sense to me that the original value should be accessible as `super` in these methods, but I don't think dry-struct actually lets us work like that. Since it defines `attr_reader` methods directly on the class, and not in some included module, we end up having to refer to the original values as `@attribute_name`, which I think is not ideal.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ COMPONENTS.each do |component|
 end
 
 gem 'dry-types', git: 'https://github.com/dry-rb/dry-types', branch: 'master'
+gem 'dry-struct', git: 'https://github.com/dry-rb/dry-struct', branch: 'skip-overridding-attr-readers'
 
 group :sql do
   gem 'sequel', '~> 4.45'
--- a/core/spec/unit/rom/relation/struct_namespace_spec.rb
+++ b/core/spec/unit/rom/relation/struct_namespace_spec.rb
@@ -136,7 +136,7 @@ RSpec.describe ROM::Relation, '#struct_namespace' do
       end
 
       it 'gives access to non-attribute methods defined in the struct superclass' do
-        expect(admins.first.shared_user?).to be_true
+        expect(admins.first).to be_shared_user
       end
 
       it 'gives access to overridden attribute methods idefined n the struct superclass' do
--- a/core/spec/unit/rom/relation/struct_namespace_spec.rb
+++ b/core/spec/unit/rom/relation/struct_namespace_spec.rb
@@ -2,12 +2,28 @@ require 'rom/relation'
 
 RSpec.describe ROM::Relation, '#struct_namespace' do
   let(:dataset) do
-    [{ id: 1, name: 'Jane' }]
+    [{ id: 1, name: 'Jane', age: 30 }]
   end
 
   before do
-    module Test::Entities
+    module Test::BaseEntities
       class User < ROM::Struct
+        def name
+          "Not Jane"
+        end
+
+        def age
+          "#{super} years old"
+        end
+
+        def shared_user?
+          true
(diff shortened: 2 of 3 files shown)


timriley:
Pinging @flash-gordon on this one, on @solnic's suggestion )

flash-gordon:
Say we have a base class for user structs
```ruby
module Entities
  class User < ROM::Struct
    def age
      "#{ super } years old"
    end
  end
end
```

We want that `Entities::User#age` method defined on top of attribute readers. We need to put readers in a module then and include it in `Entities::User`. So far so good, but there is a downside, all the readers will be available in the base class including ones defined for aggregates which is not ideal.

We can rebind the methods defined in `Entities::User` to all its successors, fortunately, Ruby allows this trick

```ruby
class AutogeneratedUserClass < Entities::User
  define_method(:age, superclass.instance_method(:age))
end
```

It works like this
```ruby
class A
  def age
    30
  end
end

class B < A
  def age
    "#{ super } years old"
  end
end

class C < B
  define_method(:age, superclass.instance_method(:age))
end

puts C.new.age #=> 30 years old years old
```

Combined with attribute readers defined in a separate module we'll be able you call attributes via `super`.

Things are becoming much more complex when inheritance comes into play. What if `Entites::User` wasn't a direct descendant of `ROM::Struct`? TL;DR I suggest to prohibit indirect inheritance and use modules instead in such cases. In contrast to inheritance, we can easily track modules inclusion with a simple overloading of the `include` method and include them only in subclasses.

In practice, this means a custom struct class becomes an outline of a real class, all instance methods will be rebound to a subclass and all modules will be included/prepended in a subclass. 

Initially, this all looked like a mess to me, then I gave it another thought. 
You can check out the PoC https://github.com/rom-rb/rom/compare/rom-struct-overridden-attribute-defects-fix?expand=1

solnic:
Can this be closed?

flash-gordon:
@solnic dunno, it doesn't look we're going to do something with this yet

solnic:
Given that we're moving to setting attributes inside `@attributes`, we can simply introduce a `get` or `read` method for reading values and call it a day.

solnic:
Closing because I assume it's clear we're not going to do anything about it. Let me know if you think otherwise.

PR #455, opened 2017-10-06
Use dry-types HEAD

alyssais:
This is required since rom-rb/rom-sql@4746e546dee81895e9c88c6fad44f047316730ff increased the minimum required dry-types version for rom-sql to one that hasn't been released yet.


solnic:
Thanks :)

PR #456, opened 2017-10-06
Include Enumerable in ROM::Relation::Materializable

alyssais:
This means that `#each` doesn't work differently to other `Enumerable` methods like `#map`, which are currently forwarded to the relation.

The current behavior is very surprising: `#each` will yield model objects, while `#map` and other `Enumerable` methods will yield hashes.

Since `Enumerable` provides a `#find` method, a spec that relied on `Materializable` not having a `#find` had to be updated to use a different non-existent method.

Additionally, the definition for `ROM::Relation::Materializable#first` can be removed since `#first` is provided by `Enumerable`.

### DIFF
--- a/core/lib/rom/relation/materializable.rb
+++ b/core/lib/rom/relation/materializable.rb
@@ -4,6 +4,8 @@ module ROM
     #
     # @api public
     module Materializable
+      include Enumerable
+
       # Coerce the relation to an array
       #
       # @return [Array]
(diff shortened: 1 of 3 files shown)


solnic:
Hey, thanks for addressing this. I'm aware of "the map problem" unfortunately we can't fix it by including entire `Enumerable` there. If we do this, all relations, across all adapters, will have enumerable interface. It's something we may consider in rom 5.0.0 but now it's too late.

What we could do instead, is defining our own `#map` in `rom-sql`. I'm going to close this because we can't do it as part of 4.0 release.

Issue #436, opened 2017-07-04
rom-repository doesn't handle whitespaces in attribute names


wrozka:
I've noticed that `rom-repository` doesn't handle whitespaces in attribute names while working with a mssql database. 

Here is my schema:
```sql
CREATE TABLE dbo.bugs
	( id          int             NOT NULL
	, white space nvarchar(max)        NULL
	)
```

the script:
```rb
require 'rom-sql'
require 'rom-repository'

container = ROM.container(:sql, "tinytds://foo:bar@localhost/bug") do |config|
  config.relation(:bugs)
end

class BugsRepository < ROM::Repository[:bugs]
end

container.relations[:bugs].delete
container.relations[:bugs].insert("id" => 1, "white space" => "WTF")

BugsRepository.new(container).bugs.to_a
```

and the stacktrace:
```
/Users/wrozka/.rvm/gems/ruby-2.4.1/gems/rom-mapper-1.0.0.beta1/lib/rom/open_struct.rb:13:in `instance_variable_set': `@white space' is not allowed as an instance variable name (NameError)
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/rom-mapper-1.0.0.beta1/lib/rom/open_struct.rb:13:in `block in initialize'
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/rom-mapper-1.0.0.beta1/lib/rom/open_struct.rb:12:in `each'
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/rom-mapper-1.0.0.beta1/lib/rom/open_struct.rb:12:in `initialize'
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/transproc-1.0.2/lib/transproc/class.rb:30:in `new'
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/transproc-1.0.2/lib/transproc/class.rb:30:in `constructor_inject'
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/transproc-1.0.2/lib/transproc/function.rb:47:in `call'
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/transproc-1.0.2/lib/transproc/function.rb:47:in `call'
	from /Users/wrozka/.rvm/gems/ruby-2.4.1/gems/transproc-1.0.2/lib/transproc/array.rb:42:in `block in map_array'
[...]

PR #461, opened 2017-10-23
Raise error when invalid attribute

GustavoCaso:
Solves #436 

Raises an error when invalid attribute name is defined by the SQL schema.

I did not know if this is the best place for raising this error. While I get some info I'm also looking for different places to raise such exception

### DIFF
--- a/core/lib/rom/schema.rb
+++ b/core/lib/rom/schema.rb
@@ -62,6 +62,8 @@ module ROM
 
     DEFAULT_INFERRER = Inferrer.new(enabled: false).freeze
 
+    VALID_IVAR_REGEX = /^[a-z_][a-zA-Z_0-9]*$/
+
     extend Initializer
 
     include Dry::Equalizer(:name, :attributes, :associations)
@@ -360,7 +362,12 @@ module ROM
     #
     # @api private
     def finalize_attributes!(gateway: nil, relations: nil)
-      inferrer.(self, gateway).each { |key, value| set!(key, value) }
+      inferrer.(self, gateway).each do |key, value|
+        value.each do |attr|
+          raise MapperInvalidAttributeName.new(attr.name) unless attr.name.match(VALID_IVAR_REGEX)
+        end
+        set!(key, value)
+      end
 
       yield if block_given?
 
--- a/core/lib/rom/schema.rb
+++ b/core/lib/rom/schema.rb
@@ -343,6 +343,17 @@ module ROM
       self.equal?(canonical)
     end
 
+    # Return if a schema attribute is valid
+    #
+    # @param [Attribute] attribute
+    #
+    # @return [Boolean]
+    #
+    # @api public
+    def valid?(attribute)
+      !!attribute.name.match(VALID_IVAR_REGEX)
+    end
+
     # Finalize a schema
     #
     # @return [self]
--- a/core/lib/rom/schema.rb
+++ b/core/lib/rom/schema.rb
@@ -363,8 +363,10 @@ module ROM
     # @api private
     def finalize_attributes!(gateway: nil, relations: nil)
       inferrer.(self, gateway).each do |key, value|
-        value.each do |attr|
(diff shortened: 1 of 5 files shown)


GustavoCaso:
This is the different configurations I used to reproduce the error

```ruby
config = ROM::Configuration.new(:sql, "postgresql://localhost/whitespace")


class Bugs < ROM::Relation[:sql]
  schema(infer: true)

  def index
    select { id.as(:uid) }
  end
end

class BugsRepository < ROM::Repository[:bugs]
end

config.register_relation Bugs

container = ROM.container(config)
connection = container.gateways[:default].connection
connection.execute "DELETE FROM bugs"
connection.execute "INSERT INTO bugs VALUES (1, 'WTF')"

repo = BugsRepository.new(container)

puts repo.bugs.to_a
```

```ruby
container = ROM.container(:sql, "postgresql://localhost/whitespace") do |config|
  config.relation(:bugs)
end

class BugsRepository < ROM::Repository[:bugs]
end

container.relations[:bugs].delete
container.relations[:bugs].insert("id" => 1, "white space" => "WTF")

BugsRepository.new(container).bugs.to_a
```

solnic:
I believe this can be simply handled in `Schema#finalize_attributes`

GustavoCaso:
You are right, after getting my hand in the finalize process and got to the same conclusion.

I will update the code.

Thanks for the hint.

GustavoCaso:
@solnic after checking the code and making the test pass. I'm thinking maybe we could move the check to `Inferrer#call` https://github.com/rom-rb/rom/blob/master/core/lib/rom/schema/inferrer.rb#L48 so we remove the `if` from the code

GustavoCaso:
@solnic do you like this approach better?

solnic:
Thanks! I'm gonna merge it in but then I'll follow up in master with some minor tweaks.

GustavoCaso:
@solnic the work that is missing is related to users using `alias`?

GustavoCaso:
@solnic should we continue working on this one? If you point me to the work that is missing I would love to finish with it.

solnic:
@GustavoCaso 👋 OK let's try this one more time 😉 to fix this, all we need to do is to raise the error [right here](https://github.com/rom-rb/rom/blob/master/core/lib/rom/struct_compiler.rb#L73-L74) when `name` includes a white-space.

core/lib/rom/schema.rb
```diff
@@ -62,6 +62,8 @@ class Schema
 
     DEFAULT_INFERRER = Inferrer.new(enabled: false).freeze
 
+    VALID_IVAR_REGEX = /^[a-z_][a-zA-Z_0-9]*$/
```
gmkohler: Why must the instance variable have a lowercase letter first?  This seems to work fine:

```rb
2.6.3 :001 > Object.new.instance_variable_set :@A, :foo
 => :foo
2.6.3 :002 > "A".match( /^[a-z_][a-zA-Z_0-9]*$/)
 => nil
```

Would it make sense to catch the `NameError` raised by Ruby instead?
```rb
2.6.3 :005 > Object.new.instance_variable_set :"@A bad choice", :foo
Traceback (most recent call last):
        6: from /Users/gregorykohler/.rvm/rubies/ruby-2.6.3/bin/irb:23:in `<main>'
        5: from /Users/gregorykohler/.rvm/rubies/ruby-2.6.3/bin/irb:23:in `load'
        4: from /Users/gregorykohler/.rvm/rubies/ruby-2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        3: from (irb):5
        2: from (irb):5:in `rescue in irb_binding'
        1: from (irb):5:in `instance_variable_set'
NameError (`@A bad choice' is not allowed as an instance variable name)
```

(I haven't found where in the C code the validity is defined)

solnic:
this needs to be reworked, so I'm closing this PR

Issue #460, opened 2017-10-21
Schema's :timestamp plugin broken with rom-sql


cmavromoustakos:
I recently upgraded to: `rom-core-4.0.0` and `rom-sql-2.0.0` from `rom (3.3.0)` and `rom-sql (1.3.3)` and started getting the error show below. 

The line that seems to be causing the exception is:

```
 class Persistence::Relations::Users < ROM::Relation[:sql]
 ┆ schema(:identity_providers, infer: true) do
 ┆ ┆ use :timestamps
 ┆ end

```

```
NoMethodError: undefined method `indexed?' for #<ROM::Attribute:0x007fbed5e9a5e8>
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-core-4.0.0/lib/rom/attribute.rb:423:in `method_missing'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-sql-2.0.0/lib/rom/sql/schema/inferrer.rb:122:in `mark_indexed'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-sql-2.0.0/lib/rom/sql/schema/inferrer.rb:51:in `block in infer_from_database'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-sql-2.0.0/lib/rom/sql/schema/inferrer.rb:51:in `map'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-sql-2.0.0/lib/rom/sql/schema/inferrer.rb:51:in `infer_from_database'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-sql-2.0.0/lib/rom/sql/schema/inferrer.rb:35:in `call'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-core-4.0.0/lib/rom/schema.rb:361:in `finalize_attributes!'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-sql-2.0.0/lib/rom/sql/schema.rb:146:in `finalize_attributes!'
/Users/cmavromoustakos/.rvm/gems/ruby-2.4.1/gems/rom-core-4.0.0/lib/rom/schema.rb:56:in `block in <class:Schema>'
[...]

PR #464, opened 2017-10-24
Fix ROM::Schema.define method

v-kolesnikov:
I think `Schema.definition` already worked correctly before, until we lost `attr_class` agrument for schema initilazer [here](https://github.com/rom-rb/rom/commit/c7518fc28#diff-3ac454f9c196d0f7c33922c20f51583cL117).
It also is reason of #460


solnic:
Hmm yeah, this LGTM. @flash-gordon ?

flash-gordon:
cool, thank you!

PR #465, opened 2017-10-29
[core] set relation registry in schema upon finalization

GustavoCaso:
@solnic @flash-gordon 

I created this PR so we can discuss this https://github.com/rom-rb/rom-sql/issues/247. The finalize process still quite complex and I'm sure I'm missing something.

For what it looks like at this point when this message `configuration.relations.registry.created` is trigger the `RealtionRegistry` has been created. 

So I thought might be a simple solution to set the relations for each schema at this point.

### DIFF
--- a/core/lib/rom/schema.rb
+++ b/core/lib/rom/schema.rb
@@ -54,6 +54,7 @@ module ROM
 
       unless schema.frozen?
         schema.finalize_attributes!(gateway: gateway, relations: registry)
+        schema.set!(:relations, registry)
       end
     end
 
@@ -416,6 +417,12 @@ module ROM
       [:schema, [name, attributes.map(&:to_ast)]]
     end
 
+    # @api private
+    def set!(key, value)
+      instance_variable_set("@#{ key }", value)
+      options[key] = value
+    end
+
     private
 
     # @api private
@@ -447,12 +454,6 @@ module ROM
         @primary_key_name = primary_key[0].meta[:name]
         @primary_key_names = primary_key.map { |type| type.meta[:name] }
       end
-    end
-
-    # @api private
-    def set!(key, value)
-      instance_variable_set("@#{ key }", value)
-      options[key] = value
     end
 
     memoize :count_index, :name_index, :source_index, :to_ast, :to_input_hash, :to_output_hash
--- a/core/spec/integration/setup_spec.rb
+++ b/core/spec/integration/setup_spec.rb
@@ -30,6 +30,13 @@ RSpec.describe 'Configuring ROM' do
       expect(tasks.users).to eql(users)
       expect(tasks.users.commands).to be_kind_of(ROM::CommandRegistry)
     end
+
+    it 'configures rom schema to store relations' do
+      users_schema = users_relation.schema
+      tasks_schema = tasks_relation.schema
+      expect(users_schema.relations[:users]).to eql(users_relation)
+      expect(tasks_schema.relations[:tasks]).to eql(tasks_relation)
+    end
   end
 
   context 'without schema' do
--- a/core/lib/rom/setup/finalize/finalize_relations.rb
+++ b/core/lib/rom/setup/finalize/finalize_relations.rb
@@ -90,7 +90,7 @@ module ROM
 
         notifications.trigger(
           'configuration.relations.schema.set',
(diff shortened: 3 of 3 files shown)


core/lib/rom/schema.rb
```diff
@@ -42,6 +42,7 @@ class Schema
       registry.each do |_, relation|
         unless relation.schema.frozen?
           relation.schema.finalize_associations!(relations: registry)
+          relation.schema.finalize_relations!(relations: registry)
```
solnic: Instead of doing this, it'd be better to [tweak this line](https://github.com/rom-rb/rom/blob/master/core/lib/rom/setup/finalize/finalize_relations.rb#L106) and do:

``` ruby
options = { __registry__: registry, mappers: mappers, schema: schema.with(relations: registry), **plugin_options }
```

core/spec/integration/setup_spec.rb
```diff
@@ -30,6 +30,13 @@
       expect(tasks.users).to eql(users)
       expect(tasks.users.commands).to be_kind_of(ROM::CommandRegistry)
     end
+
+    it 'configures rom schema to store relations' do
+      users_schema = users_relation.schema
+      tasks_schema = tasks_relation.schema
+      expect(users_schema.relations).to be_kind_of(ROM::RelationRegistry)
+      expect(tasks_schema.relations).to be_kind_of(ROM::RelationRegistry)
```
solnic: To make this spec more robust, it'd be good to do:

``` ruby
expect(users_schema.relations[:users]).to eql(users_relation)
expect(users_schema.relations[:tasks]).to eql(tasks_relation)
```

GustavoCaso:
@flash-gordon @solnic I updated the code so we can merge it and work on the rom-sql with support for accessing relations in projection/restriction blocks 😄

core/lib/rom/schema.rb
```diff
@@ -131,6 +131,8 @@ def self.attributes(attributes, attr_class)
     def initialize(*)
       super
 
+      initialize_primary_key_names
```
solnic: why was this added?

GustavoCaso: @solnic looks like after `schema.with(relations: registry)` the schema it returns have not initialized `primary_keys_names` and we were getting some errors in different parts of the specs, maybe the problem does not lie in the schema initialize method but in another part of the code.

```ruby
ROM::SQL::MissingPrimaryKeyError:
        Missing primary key for :users
```
Once I added that line the build looks like its working, but maybe there is something else that I'm missing. I happy will keep looking for other solution.

flash-gordon:
@solnic @GustavoCaso I force-pushed the branch, I made it take PK names from canonical schema, I think this should be fine for now

core/lib/rom/schema.rb
```diff
@@ -416,6 +416,22 @@ def to_ast
...
+      end
+    end
+
+    def primary_key_names
+      if canonical?
+        @primary_key_names if defined? @primary_key_names
+      else
+        canonical.primary_key_name
+      end
+    end
```
solnic: we shouldn't have this type of logic at run-time :/ How about we make these part of options instead? This way they can be maintained across instances.

core/spec/integration/setup_spec.rb
```diff
@@ -30,6 +30,13 @@
       expect(tasks.users).to eql(users)
       expect(tasks.users.commands).to be_kind_of(ROM::CommandRegistry)
     end
+
+    it 'configures rom schema to store relations' do
+      users_schema = users_relation.schema
+      tasks_schema = tasks_relation.schema
+      expect(users_schema.relations[:users]).to eql(users_relation)
+      expect(tasks_schema.relations[:tasks]).to eql(tasks_relation)
+    end
```
solnic: So, I understand that this makes sure we have finalized registry accessible via schemas, but there's also this change related to primary keys, and I'm not sure how that's related. Can we / should we have a separate spec example that checks primary key methods? ie canonical vs projected?

core/lib/rom/schema.rb
flash-gordon: couldn't we just memoize them instead? Passing them as options is also not required, strictly speaking, because they are not always needed. And yes, either memoization or passing them with options is better than using ivars

PR #468, opened 2017-11-01
[core] Remove global cache from command compiler

flash-gordon:


abrthel:
A bit late but **thank you**. This solved a really weird bug I was having in a side project.

solnic:
@abrthel good to know, thanks for letting us know :)

Issue #467, opened 2017-11-01
Can't use timestamps plugin from inside a repo


solnic:
_From @alexandru-calinoiu on April 23, 2017 15:36_

My repo look like 

```
    class Entries < IngestService::Repository[:entries]
      relations :ingest_steps
      commands :create, use: :timestamps
   end
```

I expect that calling `Entries.new.create(args)` will add timestamps to my entry, but that is not the case.

The `Timestamp` plugin get's applied at runtime when I call create, only then `ClassInterface` is extended adding `timestamps` method that can be used to configure the colmuns. Without calling `timestamps` the `InstanceMethods` will never be included and thus timestamps never added.

Unfortunately my understanding of the plugins is very limited, they look like being targeted mostly to commands, will love some pointers for the right direction.

_Copied from original issue: rom-rb/rom-sql#189_

solnic:
_From @alexandru-calinoiu on April 23, 2017 15:37_

One easy way to solve it is for the timestamp_column to have default values, but is not complete as one won't be able to configure them from a repo.

solnic:
Ah, right. I'm sorry I forgot we didn't make `timestamps` plugin work with repos yet. This will be addressed in rom 4.0 as we need to improve plugin architecture first.

solnic:
In order to make this work, we need to enable passing options for command plugins [here](https://github.com/rom-rb/rom/blob/master/core/lib/rom/commands/class_interface.rb#L112-L114) and then refactor [timestamp](https://github.com/rom-rb/rom-sql/blob/master/lib/rom/sql/plugin/timestamps.rb) plugin, so that `Timestamp` is a `Module` subclass, and accepts options. Then based on these options we can automatically configure a command. I would also say that this plugin should be moved to core, as it is not sql-specific.

GustavoCaso:
@solnic I would love to help with this one

solnic:
Done via https://github.com/rom-rb/rom/commit/962bb273931a7b224bfb0c4eb52e940f76af7e40

PR #469, opened 2017-11-10
Allow options for command timestamps plugin

GustavoCaso:
Solves #467 

- Allow to pass options to command plugins
- Refactor Command Timestamp plugin to accept options
- Move Command Timestamp plugin to rom-core

After this changes the `timestamp` plugin will works as before and also accepts initial options to set timestamps.

```ruby
class CreateTask < ROM::Commands::Create[:sql]
  result :one
  use :timestamps, timestamps: %i(created_at, updated_at), datestamps: %i(:written)
end
```

Also in order to been able to set a fully working command from a `Relation` we need to allow passing options to the `commands` API, since we support setting multiple commands, I decided to go with the approach of having a new argument `plugins_options` been a hash where each key must be equal to the plugins name.

```ruby
class UserRepo < ROM::Repository[:users]
  commands :create, use: %i(schema timestamps), plugins_options: { timestamps: datestamps: %i(:written) }
end
```

### Missing 
- [x] Modify `Repository#commands` to accepts options for plugin
- [ ] Remove `Timestamp` from `rom-sql` https://github.com/rom-rb/rom-sql/pull/264

### DIFF
--- a/core/lib/rom/command_compiler.rb
+++ b/core/lib/rom/command_compiler.rb
@@ -52,6 +52,10 @@ module ROM
     #   @return [Array<Symbol>] a list of optional plugins that will be enabled for commands
     option :plugins, optional: true, default: -> { EMPTY_ARRAY }
 
+    # @!attribute [r] plugins_options
+    #   @return [Hash] a hash of options for the plugins
+    option :plugins_options, optional: true, default: -> { EMPTY_HASH }
+
     # @!attribute [r] meta
     #   @return [Array<Symbol>] Meta data for a command
     option :meta, optional: true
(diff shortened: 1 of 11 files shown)


amarshall:
This was released in rom v4.1.0 and rom-sql 2.3.0 on 2017-11-17, however, it is broken as rom-repository did not receive a new release with [its changes from this patch](https://github.com/rom-rb/rom/pull/469/files#diff-ad522966fa1c6b24df786ab56617dab8), resulting in `plugins_options` not being passed through when using the released versions.

It seems rom-repository release on that date was missed—would it be possible to get a release of rom-repository with these changes?

solnic:
@amarshall oh 💩 😓 thanks for reporting this. I'll push new rom-repo today.

solnic:
@amarshall I've released rom 4.1.1 which will pull in rom-repository 2.0.2, and this one supports `plugins_options`

PR #476, opened 2017-12-28
Make timestamp plugin work with changesets

### LINKED ISSUE #475: timestamp plugin breaks with changesets
Using a changeset in combination with a custom command using the `timestamps` plugin results in an error:

```ruby
require 'bundler/inline'

gemfile(install: true) do
  source "http://rubygems.org"

  gem 'rom'
  gem 'rom-sql'
  gem 'sqlite3'
end


class Users < ROM::Relation[:sql]
  schema(infer: true)
end

  relation :users
  register_as :create

  use :timestamps
  timestamp :created_at, :updated_at
end

rom = ROM.container(:sql, "sqlite::memory") do |conf|

  conf.default.create_table :users do
    primary_key :id
    column :name, String, null: false
    column :created_at, DateTime, null: false
    column :updated_at, DateTime, null: false
  end

  conf.register_relation(Users)
  conf.register_command(CreateUser)

end

rom.commands[:users].create.call(name: 'Test User one')

puts rom.relations[:users].to_a

rom.relations[:users].changeset(:create, name: 'Test User two').commit

puts rom.relations[:users].to_a
```

results in: 
```
{:id=>1, :name=>"Test User one", :created_at=>2017-12-28 18:48:28 -0500, :updated_at=>2017-12-28 18:48:28 -0500}
/home/flip/.gem/ruby/2.4.1/gems/rom-sql-2.3.0/lib/rom/sql/commands/create.rb:22:in `execute': wrong number of arguments (given 0, expected 1) (ArgumentError)
	from /home/flip/.gem/ruby/2.4.1/gems/rom-core-4.1.0/lib/rom/command.rb:280:in `call'
	from /home/flip/.gem/ruby/2.4.1/gems/rom-sql-2.3.0/lib/rom/sql/commands/error_wrapper.rb:16:in `call'
	from /home/flip/.gem/ruby/2.4.1/gems/rom-core-4.1.0/lib/rom/commands/composite.rb:17:in `call'
	from /home/flip/.gem/ruby/2.4.1/gems/rom-changeset-1.0.1/lib/rom/changeset/stateful.rb:201:in `commit'
	from ./rom-repo.rb:42:in `<main>'
```

cflipse:
Changesets pass themselves as an argument to command pipelines, rather
than the raw tuple.  Because the timestamp plugin was using Hash/Array
to determine the method to use for blending new data, it was triggering
a failure when the changeset fell through the case statement here.

We should probably keep an eye out for any similar patterns in command
hooks as the resulting error message was an utterly unclear:
      ArgumentError:
       wrong number of arguments (given 0, expected 1)
     # /home/flip/.gem/ruby/2.4.2/gems/rom-sql-2.3.0/lib/rom/sql/commands/create.rb:22:in `execute'

fixes #475

### DIFF
--- a/core/lib/rom/plugins/command/timestamps.rb
+++ b/core/lib/rom/plugins/command/timestamps.rb
@@ -70,12 +70,7 @@ module ROM
           def set_timestamps(tuples, *)
             timestamps = build_timestamps
 
-            case tuples
-            when Hash
-              timestamps.merge(tuples)
-            when Array
-              tuples.map { |t| timestamps.merge(t) }
-            end
+            map_input_tuples(tuples) { |t| timestamps.merge(t) }
           end
 
           private
--- a/changeset/spec/integration/changeset_spec.rb
+++ b/changeset/spec/integration/changeset_spec.rb
@@ -36,6 +36,25 @@ RSpec.describe 'Using changesets' do
       expect(delete).to be_kind_of(ROM::Changeset::Delete)
     end
 
+    it 'works with command plugins' do
+      configuration.commands(:books) do
+        define(:create) do
+          use :timestamps
+          timestamp :created_at, :updated_at
+          result :one
+        end
+      end
+
+      changeset = books.changeset(:create, title: "rom-rb is awesome")
+
+      result = changeset.commit
+
+      expect(result.id).to_not be(nil)
+      expect(result.title).to eql("rom-rb is awesome")
+      expect(result.created_at).to be_instance_of(Time)
+      expect(result.updated_at).to be_instance_of(Time)
+    end
+
     it 'can be passed to a command' do
       changeset = users.changeset(:create, name: "Jane Doe")
       command = users.command(:create)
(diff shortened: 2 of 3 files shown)


core/lib/rom/plugins/command/timestamps.rb
```diff
@@ -70,10 +70,9 @@ def datestamp_columns
           def set_timestamps(tuples, *)
             timestamps = build_timestamps
 
-            case tuples
-            when Hash
+            if tuples.respond_to?(:merge)
```
cored: @cflipse Just curious, would this snippet be clearer? 
```ruby
timestamps.merge(tuples) if tuples.respond_to?(:merge)
tuples.map { |t| timestamps.merge(t) } if tuples.respond_to?(:map)
```

cflipse: responding to one or the other method are not mutually exclusive.

solnic: there's a method called `Command#with_input_tuples` which yields hashes, IIRC it should be available in this context

abrthel: If I remember right, `Command#with_input_tuples` is a private method defined inside the create & update commands in rom-sql.

solnic: Bummer. I’d say it should be moved to rom-core then :)

abrthel: 👍 plus it has the added benefit of removing the duplication between the rom-sql create and update commands.

cflipse: Just did a trial run of yanking that over.

One downside:  The method as-is transforms a single-hash tuple into an array of tuples ... which breaks a few specs, because they (rightly) assume the airity of the tuples remains consistant

solnic: @cflipse gah right, well then we need something that does `with_input_tuples` + returns either one or many results :/ Remember when I said I regret adding `result` option to commands? :D This is precisely how it complicates commands.

cflipse: it's worse than that, actually:

the mapping / result needs to agree with it's argument.  `update`  command,  for example can generally be considered to return `:many` but may only be called with a single input tuple.

I can get specs to pass by checking on the hashiness of the input tuple, but I suspect this might end up causing problems with pipelines ... though maybe those are supposed to be treated as more array-like anyway?

cflipse: Aaaand, I can't use the `with_input_tuples` name because rom-sql has already internalized the idea that it always returns an array, and so it breaks multi-insert.  

`map_input_tuples` it is.

PR #478, opened 2018-01-16
WIP Use dry-types/dry-struct from master

flash-gordon:
@solnic I wonder what's our plan for default types. It's easy to make them work as before (i.e. `nil` triggers a default value) but I think we should change this behavior in the next version. FYI SQL's defaults work in the same way, if you insert `NULL` it doesn't use the default value.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,8 @@ COMPONENTS.each do |component|
   gem "rom-#{component}", path: Pathname(__dir__).join(component).realpath
 end
 
-gem 'dry-struct'
+gem 'dry-types', git: 'http://github.com/dry-rb/dry-types'
+gem 'dry-struct', git: 'http://github.com/dry-rb/dry-struct'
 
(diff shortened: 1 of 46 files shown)


solnic:
To be honest, this should be adapter-specific. ie if a database supports default values, then it should be used. Using default values in structs that rom loads is a code smell for me.

flash-gordon:
@solnic it's more about input schemas that inject a default value on `nil`. I would argue they should do it on missing key instead.

solnic:
@flash-gordon ah right, well, would be good to maintain treating `nil` as a trigger for default value. Otherwise we'd have to rewrite input hashes every time a command is applied, which may have negative performance impact.

flash-gordon:
@solnic hm, actually, I can't see what's the difference between these approaches, even `nil`s have to be overridden with default values, hashes will be rewritten in both cases

PR #479, opened 2018-02-03
Look up the most nested constant first in the CustomNamespace strategy

flash-gordon:
In my app, I have `App::Events` which is a module not related to the persistence layer and I also have `App::Relations::Events` which is a relations class. `CustomNamespace` wrongfully picks up `App::Events` rather than App::Relations::Events. This PR fixes the issue, now it starts with the longest constant name.

### DIFF
--- a/core/lib/rom/setup/auto_registration_strategies/custom_namespace.rb
+++ b/core/lib/rom/setup/auto_registration_strategies/custom_namespace.rb
@@ -23,13 +23,14 @@ module ROM
       #
       # @api private
       def call
-        potential = []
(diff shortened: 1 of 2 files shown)


flash-gordon:
I had some failures on jRuby. Initially, I tried to bump its version but it doesn't work well because of I tried https://github.com/jruby/jruby/issues/4895. Then I updated specs to forcefully "unload" constants in https://github.com/rom-rb/rom/pull/479/commits/bad37407616e3d394335b823f06807741a1099ca, this slowed down specs by 0.5s (2.2 -> 2.7) or so but I believe a clean environment is better than speed here, we can tag these specs as slow if needed.

PR #480, opened 2018-02-15
Provide adapter to schema DSL instance

alsemyonov:
Fixes ":plugin doesn't exist in ROM::AdapterPluginRegistry registry"
when using adapter-specific plugins.

### DIFF
--- a/core/lib/rom/relation/class_interface.rb
+++ b/core/lib/rom/relation/class_interface.rb
@@ -105,6 +105,7 @@ module ROM
           @schema_proc = proc do |*args, &inner_block|
             schema_dsl.new(
               relation_name,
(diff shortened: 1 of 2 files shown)


solnic:
It would be good to write a spec, so that nobody breaks it by an accident in the future.

alsemyonov:
@solnic, please check the spec. Is it good enough?

solnic:
Yes perfecto :) Thank you for addressing this! I'll push a bug-fix release tomorrow.

alsemyonov:
Cool! Thank you.

solnic:
@alsemyonov looks like this broke `spec/integration/setup_spec.rb` :( could you take a look?

PR #484, opened 2018-03-26
Allow specifying timestamps manually when using mappers

cgeorgii:
There was no way to override timestamps when using the `:add_timestamps` and `:touch` [transformations](http://rom-rb.org/4.0/learn/core/changesets/#changeset-mapping) with changesets.

I discussed it with @flash-gordon and decided to open a PR to add the possibility. I added tests and tried to follow the conventions, but let me know if anything is missing.

Edit: opened a [pull request](https://github.com/rom-rb/rom-rb.org/pull/255) to rom-rb.org as well.

### DIFF
--- a/changeset/lib/rom/changeset/pipe.rb
+++ b/changeset/lib/rom/changeset/pipe.rb
@@ -14,11 +14,11 @@ module ROM
 
       def self.add_timestamps(data)
         now = Time.now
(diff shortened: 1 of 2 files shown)


flash-gordon:
@cgeorgii thanks, looking ❤️

PR #495, opened 2018-05-10
Make ROM::Inflector configurable

ianks:
I ran into an issue when building adapter that required me to configure the inflector. In order to do this, I had to override the `ROM::Inflector` const. This PR makes it so the inflector can be configured. PS: A few rubocop autofixes are polluting the diff; let me know if you want me to take those out!


flash-gordon:
hmm, we don't like global state => we'll have an injectable inflector in 5.0, I wonder if we need something in between in 4.0 @solnic

ianks:
@solnic Can you further specify the interface you would like to see for this? I'm happy to help but want to make sure I'm clear as to exactly what you want.

flash-gordon:
@ianks see how cache passed around https://github.com/rom-rb/rom/search?utf8=%E2%9C%93&q=cache&type=

PR #496, opened 2018-05-15
Struct compiler now checks if a class exists before creating a new one

radar:
This is related to https://github.com/dry-rb/dry-core/pull/26. I am still seeing an issue in some projects, which you can still reproduce using the steps in dry-rb/dry-core#25.

---

A class may already exist and just not be available (yet) because it hasn't been autoloaded. The StructCompiler class should call const_get in an attempt to load this class. If it can load the class, then it should use this class and not go to Dry::Core::ClassBuilder to build an entirely new class.

---

In my little script to reproduce this issue:

```ruby
repo = ProjectRepository.new(ROM.env)
project = repo.projects.to_a.first
puts project.class.to_s == "Projects::Project"
puts project.respond_to?(:to_model)
```

This shows up "false" and "true". I believe this is because the `Projects::Project` constant is not being autoloaded in `Dry::Core::ClassBuilder` due to the line I pointed out in dry-rb/dry-core#25. I am on Ruby 2.5, and so that line isn't called.

Reviewing my decision there, I think making `StructCompiler` attempt to find the class is a better course of action, as `ClassBuilder` would then only be used in the cases where we had to actually build a class. This is the route I've gone down with this PR.

With this patch applied, my little script shows up "true" and "true", indicating that the behaviour is correct... at least according to that script.

---

I couldn't work out a non-ghetto way of clearing the constant in the tests. Sorry.

### DIFF
--- a/mapper/lib/rom/struct_compiler.rb
+++ b/mapper/lib/rom/struct_compiler.rb
@@ -34,11 +34,13 @@ module ROM
         if attributes.empty?
           ROM::OpenStruct
         else
-          build_class(name, ROM::Struct, ns) do |klass|
-            attributes.each do |(attr_name, type)|
-              klass.attribute(attr_name, type)
-            end
+          klass = get_class(name, ns) || build_class(name, ROM::Struct, ns)
+
+          attributes.each do |(attr_name, type)|
+            klass.attribute(attr_name, type)
           end
+
+          klass
         end
       end
     end
@@ -93,6 +95,13 @@ module ROM
     def visit_enum(node)
       type_node, * = node
       visit(type_node)
+    end
+
+    # @api private
+    def get_class(name, ns)
+      ns.const_get(class_name(name))
+    rescue NameError
+      nil
     end
 
     # @api private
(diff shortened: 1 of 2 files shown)


radar:
@solnic Am I on the right track to solving this problem here? I just ran into it again today.

flash-gordon:
@radar no, it's wrong, I wonder why our tests are even passing 😟 
The idea is to use user-provided structs as base classes, whereas your changes modify the base class itself. In a real-world application `rom` creates more than one subclass and defines required attributes in every one of them. With your changes, it will try to define them all in a single struct class, this will end up with an error.

We need to improve our test suite.

solnic:
I'm going to close this as it's not the right solution. Please report an issue describing the problem and we can think about addressing it properly. My gut feeling is telling me this should be solved in rom-rails because it's related to const auto-loading and code reloading.

nfstern:
@radar:  Working thru your ExplodingRails and you said to comment here on whether the fix worked.
adding
gem 'rom-mapper', github: 'radar/rom', branch: 'build-once'
does not seem to have fixed the problem.

Issue #494, opened 2018-05-04
dry-types 0.13.0 breaks how rom-relations uses relations


joelvh:
Got errors with dry-types 0.13.0 with `map_with` calls on relations in repositories.

Locking in dry-types 0.12.2 fixes this.

```
An error occurred while loading ./spec/rom/env_spec.rb.
Failure/Error: require 'rom-repository'
NameError:
  undefined method `map_with' for class `ROM::Relation::Curried'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/relation/curried.rb:26:in `<class:Curried>'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/relation/curried.rb:19:in `<class:Relation>'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/relation/curried.rb:10:in `<module:ROM>'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/relation/curried.rb:9:in `<top (required)>'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/relation.rb:16:in `require'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/relation.rb:16:in `<top (required)>'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/core.rb:11:in `require'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-core-4.2.0/lib/rom/core.rb:11:in `<top (required)>'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-repository-2.0.2/lib/rom-repository.rb:1:in `require'
# /home/travis/.rvm/gems/ruby-2.5.1/gems/rom-repository-2.0.2/lib/rom-repository.rb:1:in `<top (required)>'
# ./lib/ruby_event_store/rom.rb:3:in `require'
# ./lib/ruby_event_store/rom.rb:3:in `<top (required)>'
# ./spec/spec_helper.rb:1:in `require'
# ./spec/spec_helper.rb:1:in `<top (required)>'
# ./spec/rom/env_spec.rb:1:in `require'
# ./spec/rom/env_spec.rb:1:in `<top (required)>'
```

[...]

PR #498, opened 2018-05-23
Upgrade dry types core

yuszuv:
This updates the gem to the new `dry-types` types.

As mentioned in #494 this can't be released as `4.x`, but, however, maybe someone finds this interesting.

### DIFF
--- a/core/lib/rom/attribute.rb
+++ b/core/lib/rom/attribute.rb
@@ -24,8 +24,8 @@ module ROM
     param :type
 
     # @api private
-    def [](input)
-      type[input]
+    def [](*args)
+      type[*args]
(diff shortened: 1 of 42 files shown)


solnic:
Thanks for this. Something we could do would be to cut off a stable 4.x branch from master for the ongoing maintenance, then continue in master with 5.x effort, and this PR could be the first step. We can even consider releasing a very early 5.0.0.alpha right away, for the impatient ones who would like to start using latest dry-types together with rom. WDYT @flash-gordon & @timriley?

flash-gordon:
I cherry-picked almost all of the commits, thanks!

PR #504, opened 2018-10-05
Support passing container to Repository as either param or option

timriley:
Being able to pass `container` as an option means Repository can work with an all-options constructor signature, which makes it easier to inject additional app-specific dependencies to subclasses using tools like dry-auto_inject (with its standard kwargs strategy).

This change is backwards compatible because the standard param continues to be respected.

So now you can initialize a repo in both these ways:

```ruby
UserRepo.new(rom)
UserRepo.new(container: rom)
```

And here's an example of how it'd with with dry-auto_inject:

```ruby
module MyApp
  class Repo < ROM::Repository
    include Import[container: "persistence.rom"]
  end

  class UserRepo < Repo
    include Import["another_dep"]
    
    # container is injected via superclass
    # can also work with `another_dep` inside instance methods here
  end
end
```

@solnic @flash-gordon How's this look to you?

### DIFF
--- a/repository/lib/rom/repository.rb
+++ b/repository/lib/rom/repository.rb
@@ -86,7 +86,7 @@ module ROM
 
     # @!attribute [r] container
     #   @return [ROM::Container] The container used to set up a repo
-    param :container, allow: ROM::Container
+    option :container, allow: ROM::Container
 
     # @!attribute [r] struct_namespace
     #   @return [Module,Class] The namespace for auto-generated structs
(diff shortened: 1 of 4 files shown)


timriley:
I'm not sure what our policy with codeclimate issues is here, but I'd be happy to shorten those yardoc lines 🤷‍♂️

solnic:
> I'm not sure what our policy with codeclimate issues is here

It's to use common sense 😄

PR #510, opened 2018-12-16
Remove `Repository::Root#aggregate` method

v-kolesnikov:
Remove `Repository::Root#aggregate` in favor of `Relation#combine`

Upgrade example:

ROM 4.x

    class Users < ROM::Repository::Root[:user]
      def list
        aggregate(tasks: :tags).to_a
      end
    end

ROM 5.x

    class Users < ROM::Repository::Root[:user]
      def list
        users
          .combine(tasks: :tags)
          .to_a
      end
    end


timriley:
This looks good to me, thanks @v-kolesnikov. I'll let @solnic or @flash-gordon have final say before merging in this case, though.

If I could offer any suggestion here, it's that I'd want to put a description for the empty `it` example inside the "with nested nodes" spec example, just for consistency :)

v-kolesnikov:
@timriley Thank you for the review!

repository/spec/integration/combine_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  context 'when invalid relation name was passed to `node` method' do
+    let(:message) { ':poztz is not a valid aggregate node name' }
+
+    it 'raises arg error ' do
+      expect { combined.node(:poztz) {} }.to raise_error(ArgumentError, message)
+    end
+  end
+end
```
solnic: was this spec ported from some other place or did you write it from scratch?

v-kolesnikov: Yes, it was ported from repository/spec/integration/repository/aggregate_spec.rb https://github.com/rom-rb/rom/pull/510/files#diff-c2d607da524fcf9a023a690743348d4cL52

Is something wrong with it?

v-kolesnikov:
> @v-kolesnikov thanks, could you also add a deprecation warning that this method goes away to `release-4.2` branch?

👌

PR #512, opened 2018-12-20
Decouple attribute alias from type definition

waiting-for-dev:
This PR is created because of this thread in the forums:

https://discourse.rom-rb.org/t/ability-to-alias-an-attribute-but-keeping-type-inferred/289

Basically, it decouples an attribute alias from its type, so it will be
possible to define an alias but still make use of schema inference.

This is still a WIP, but I'm looking for some feedback in order not to do extra work.

Some considerations I'd like to have your thoughts:

- The DSL is still not adapted. I haven't tested it, but I guess it won't work
  defining an `attribute` without specifying its type, because the main purpose
  of this PR is to be able to do a simple:

  `attribute :foo, alias: :bar`
  
- The option to define an alias is called `alias:`. This is because `as:` can't
  be freely used because it is already a method name and `dry-initializer`
  creates methods from option names. We could use method renaming feature in
  `dry-initializer` for that, but then we have to juggle with the renaming in
  the rest of code, like in `aliased` and `meta` methods, and it is a bit ugly.
  The handicap is that in other places of the API `as:` is used for similar
  purposes, like in association aliasing.
  
- `alias` method has been removed because it is defined out of the box by
  `dry-initializer`.
  
- The attribute `ast` will now always return `alias: nil` when an alias is not
  defined. This is different from before, where `alias` key was not present in
  those situations. In my opinion, it makes more sense to always return all the keys.
  
- The test for returning aliased attribute identified by canonincal name has
  been removed, because I think that, being now an option and not part of the
  type, it makes no sense.

- I have redefined `ROM::SQL::Function#name` method in the repository
  integration tests. This should be changed in `rom-sql` project. How would yo
  go with it?

- I guess that it would also make sense to move `wrapped` from the `meta` to an
  `option`. Would you like to change it also in this PR?

- What do you think about `primary_key` meta? Should we consider it part of the
  type (`meta`) or also an option? Even if it related with the value (the value
  being unique), it is so only considering the rest of values. It is like
  dynamic typing. Anyway, it would be more convenient to have it as option, I
  guess.

Thanks!


solnic:
> The test for returning aliased attribute identified by canonincal name has
been removed, because I think that, being now an option and not part of the
type, it makes no sense.

Could you elaborate a bit more here?

waiting-for-dev:
> > The test for returning aliased attribute identified by canonincal name has
> > been removed, because I think that, being now an option and not part of the
> > type, it makes no sense.
> 
> Could you elaborate a bit more here?

Well, I guess this test existed in order to be sure than aliasing an attribute didn't change the `name` in the `meta` of the type. But now they are two different concepts. One thing is the type of an attribute and another thing are their options (like `alias:`). So, I think it makes no sense to test that the change in the type doesn't affect the options... why should it? I think that keeping the test just reflects the development history (the fact that the alias was part of the type before). Furthermore, `rename` seems very well tested in [`rename_spec`](https://github.com/rom-rb/rom/blob/master/core/spec/unit/rom/schema/rename_spec.rb) file.

But maybe I'm missing something :slightly_smiling_face:

solnic:
> Well, I guess this test existed in order to be sure than aliasing an attribute didn't change the name in the meta of the type.

This test exists because it checks we can still access attributes using canonical names, and `rename_spec` doesn't cover this scenario, so it seems like we should not remove the test.

waiting-for-dev:
Makes sense. I readded the test, fixed rubocop issues and forced the push rebasing from master.

There is still more to review, however. Besides the points I mentioned in the first comment, we should also decide about the `#inspect` output for `Attribute` now that it has more options.

solnic:
> Besides the points I mentioned in the first comment, we should also decide about the #inspect output for Attribute now that it has more options.

We could do:

``` ruby
def inspect
  %(#<#{self.class}[#{type.name}] #{meta.merge(options).map { |k, v| "#{k}=#{v.inspect}" }.join(' ')}>)
end
```

core/lib/rom/attribute.rb
```diff
@@ -246,7 +228,7 @@ def alias
     #
     # @api public
     def aliased(name)
-      meta(alias: name)
+      self.class.new(type, options.merge(alias: name))
     end
```
solnic: We have a shortcut for this:

``` ruby
with(alias: name)
```

core/lib/rom/attribute.rb
```diff
@@ -403,18 +385,18 @@ def to_ast
...
+      [:attribute, [name, to_read_type.to_ast(meta: false), meta_options_ast]]
     end
 
     # @api private
-    def meta_ast
-      meta_keys = %i(wrapped alias primary_key)
-      ast = meta.select { |k, _| meta_keys.include?(k) }
+    def meta_options_ast
+      keys = %i[wrapped primary_key alias]
+      ast = (meta.merge options).select { |k, _| keys.include?(k) }
```
solnic: Minor style issue: `(meta.merge(options)).select ...`

core/lib/rom/attribute.rb
```diff
@@ -23,6 +23,11 @@ class Attribute
     #   @return [Dry::Types::Definition, Dry::Types::Sum, Dry::Types::Constrained] The attribute's type object
     param :type
 
+    # @!attribute [r] type
+    #   @return [Dry::Types::String, Dry::Types::Symbol, nil] Alias to
```
solnic: This should be `@return [Symbol, nil]` (so use actual core ruby types)

core/lib/rom/attribute.rb
```diff
@@ -23,6 +23,11 @@ class Attribute
...
+    #   use instead of attribute name
+    option :alias, optional: true
```
solnic: This should be constrained by `type: Types::Strict::Symbol.optional`

waiting-for-dev:
I added a commit for each change. We can squash them later if needed.

> > Besides the points I mentioned in the first comment, we should also decide about the #inspect output for Attribute now that it has more options.
> 
> We could do:
> 
> ```ruby
> def inspect
>   %(#<#{self.class}[#{type.name}] #{meta.merge(options).map { |k, v| "#{k}=#{v.inspect}" }.join(' ')}>)
> end
> ```

As with attribute `ast`, inspect will always include `alias:` key even when the value is `nil`. Personally I prefer being explicit in these situations, but, anyway, I guess we must be consistent with the behaviour for `meta` keys (if there is still any at the end of the work in this PR).

core/lib/rom/schema.rb
```diff
@@ -138,7 +138,13 @@ def self.define(name, attributes: EMPTY_ARRAY, attr_class: Attribute, **options)
 
     # @api private
     def self.attributes(attributes, attr_class)
-      attributes.map { |type| attr_class.new(type) }
+      attributes.map do |attr|
+        if attr.is_a?(Hash)
```
waiting-for-dev: If I'm not wrong, if we want to be backward compatible with plugins, I can't just change here to use always a Hash. Instead, I'm allowing the type information to be still just the type itself.

core/lib/rom/schema/dsl.rb
```diff
@@ -72,7 +75,7 @@ def attribute(name, type, options = EMPTY_HASH)
                          "Attribute #{ name.inspect } already defined"
         end
 
-        attributes[name] = build_type(name, type, options)
+        attributes[name] = build_attribute_info(name, type, options)
```
waiting-for-dev: In order to be backward compatible with plugins redefining `#build_type`, I take an intermediate step here which forwards to `#build_type` and takes the options.

waiting-for-dev:
> Please open a PR in rom-sql which changes this, preferably with tests. You could temporary target this branch in the rom-sql's Gemfile, once it's merged we can go back to depending on master.

Before I needed to push here the update to the DSL schema. I added some comments to the code to make it clear some intentions.

core/lib/rom/schema.rb
```diff
@@ -138,7 +138,13 @@ def self.define(name, attributes: EMPTY_ARRAY, attr_class: Attribute, **options)
...
+          attr_class.new(attr[:type], **attr[:options] || {})
```
solnic: I believe this can be simplified like this:

``` ruby
 attr_class.new(attr[:type], attr.fetch(:options, EMPTY_HASH))
```

WDYT?

waiting-for-dev: Sure, much nicer. Done!

waiting-for-dev:
I'm stuck with the last change needed in order to have the new feature available.

We need the following to work:

```ruby
schema(infer: true) do
  attribute :first_name, alias: :name
end
```

Attribute inferrer checks whether an attribute has been already defined by the user [comparing names](https://github.com/rom-rb/rom/blob/master/core/lib/rom/schema/inferrer.rb#L86). However, `:name` is an option in `meta`, which is something related with the type, and with this new DSL feature we would be defining the attribute options but not the type. Looks like we should also move `:name` to be an attribute option?

solnic:
> Looks like we should also move :name to be an attribute option?

Yes, this makes perfect sense. My understanding is that first and foremost we're moving attribute-related options to the new options hash, instead of type meta. So `:name` should be moved there too.

waiting-for-dev:
https://github.com/rom-rb/rom/blob/a9a9f2d1676e2a48a1ec94c8b85e97c9c162793d/core/lib/rom/schema.rb#L140-L148

Do you think that backward compatibility issues with plugins could arise? Changing `:name` to be an option would make the attribute information extracted from the DSL to be always a Hash.

solnic:
> Changing :name to be an option would make the attribute information extracted from the DSL to be always a Hash.

Not sure if I follow, wdym by `the attribute information extracted from the DSL`?

changeset/spec/spec_helper.rb
```diff
@@ -2,6 +2,8 @@
 
 SPEC_ROOT = Pathname(__FILE__).dirname
 
+require SPEC_ROOT.join('../../spec/sql_overrides')
+
```
waiting-for-dev: I moved all the `rom-sql` overrides needed for testing to a single file. Once `rom-sq` has been adapted we just need to delete it and the lines requiring it.

core/lib/rom/attribute.rb
```diff
@@ -24,6 +24,10 @@ class Attribute
     #   @return [Dry::Types::Definition, Dry::Types::Sum, Dry::Types::Constrained] The attribute's type object
     param :type
 
+    # @!attribute [r] type
+    #   @return [Symbol] Attribute name. This is the canonical name used in the datastore.
+    option :name, type: Types::Strict::Symbol
+
```
waiting-for-dev: Being an option, `#name` method is no longer needed.

core/lib/rom/attribute.rb
```diff
@@ -312,9 +290,9 @@ def meta(opts = nil)
     #
     # @api public
     def inspect
-      opts = options.reject { |k| k == :type }
+      opts = options.reject { |k| [:type, :name].include?(k) }
       meta_and_opts = meta.merge(opts).map { |k, v| "#{k}=#{v.inspect}" }
-      %(#<#{self.class}[#{type.name}] #{meta_and_opts.join(' ')}>)
+      %(#<#{self.class}[#{type.name}] name=#{name.inspect} #{meta_and_opts.join(' ')}>)
```
waiting-for-dev: I think it makes sense to keep giving a special treatment to `name` in the `#inspect` option.

core/lib/rom/schema.rb
```diff
@@ -136,14 +136,26 @@ def self.define(name, attributes: EMPTY_ARRAY, attr_class: Attribute, **options)
...
+
     # @api private
     def self.attributes(attributes, attr_class)
       attributes.map do |attr|
-        if attr.is_a?(Hash)
-          attr_class.new(attr[:type], attr.fetch(:options, EMPTY_HASH))
-        else
-          attr_class.new(attr)
-        end
+        attr_class.new(attr[:type], attr.fetch(:options))
```
waiting-for-dev: I'll try to explain here the issue with backward compatibility. The DSL is calling `Schema.define`, which in turn is calling `Schema.attributes`. Before these changes, latter expected the attributes to be an array of types. Now it expects it to be an array of hashes `:type`/`:options`. Not sure if it is an issue with schema plugins or something.

core/lib/rom/schema.rb
```diff
@@ -136,14 +136,26 @@ def self.define(name, attributes: EMPTY_ARRAY, attr_class: Attribute, **options)
...
+    # Builds a representation of the information needed to create an
+    # attribute.
+    #
+    # This representation is consumed by `Schema.define` in order to create
+    # the actual attributes.
+    #
+    # @return [Hash] A hash with `:type` and `:options` keys.
+    #
+    # @api private
+    def self.build_attribute_info(type, options)
```
waiting-for-dev: Defining this method here, because it seems that the attribute class is decoupled from the attribute representation (`Schema.define` accepts the attribute class as argument). Otherwise, maybe we could work with `Attribute#ast`, which would be neater.

core/spec/unit/rom/attribute/method_missing_spec.rb
```diff
@@ -3,7 +3,7 @@
 
 RSpec.describe ROM::Attribute, '#method_missing' do
   subject(:attribute) do
-    ROM::Attribute.new(type)
+    ROM::Attribute.new(type, name: :foo)
```
waiting-for-dev: There are lots of places in the test suite where `define_attribute` helper could be used. However, it is not flexible enough to accept some scenarios, like a type with call to `.optional`, or a type under `ROM::Memory` namespace. Maybe a nice refactor for consistency would be whether make it more flexible or get rid off it.

spec/sql_overrides.rb
```diff
@@ -0,0 +1,60 @@
...
+    class ProjectionDSL < DSL
+      def method_missing(meth, *args, &block)
+        if schema.key?(meth)
+          schema[meth]
+        else
+          type = type(meth)
+
+          if type
+            if args.empty?
+              ::ROM::SQL::Function.new(type, name: :not_used, schema: schema)
```
waiting-for-dev: To be honest, not sure what is happening here. But seems that `name:` is not needed at all.

waiting-for-dev:
The previous commit added the end goal feature of being able to do:

```ruby
schema(infer: true) do
  attribute :name, alias: :username
end
```

Besides any change required to the API or the implementation, it would just remain to be done moving other `meta`'s to `options`, if we think it is a good idea.

spec/sql_overrides.rb
waiting-for-dev: I see. I guess that in the meantime, before facing that refactor, the only possibility is to have `:name` as optional in its `dry-initializer` definition.

core/lib/rom/attribute.rb
waiting-for-dev: Ok, I thought that maybe it was too much for an attribute. But you are right, in fact it makes no difference whether it is an attribute or a method. However, now that I give it a second view, I think the example given is incorrect:

```ruby
    # @example	
    #   class Tasks < ROM::Relation[:memory]	
    #     schema do	
    #       attribute :user_id, Types::Integer.meta(alias: :id)	
    #       attribute :name, Types::String	
    #     end	
    #   end	
    #	
    #   Users.schema[:id].name	
    #   # => :id	
    #	
    #   Users.schema[:user_id].name	
    #   # => :user_id
```

`:id` attribute is not defined in the schema, and `:user_id` name should return `:id`, isn't it? Maybe  should we just change it to:

```ruby
  # Users.schema[:name].name
  # # => :name
  #
  # Users.schema[:id].name
  # # => :id
```

?

solnic: @waiting-for-dev hah good catch, let's just stick to this:

```
    # @example	
    #   class Users < ROM::Relation[:memory]	
    #     schema(:users) do	
    #       attribute :user_id, Types::Integer, alias: :id
    #       attribute :email, Types::String
    #     end	
    #   end	
    #	
    #   users[:id].name	
    #   # => :id
    #
    #   users[:email].name	
    #   # => :email
```

core/lib/rom/schema.rb
solnic: `Schema.attributes` is a private API, no need to worry about it

core/spec/unit/rom/attribute/method_missing_spec.rb
solnic: I agree this could be improved, let's keep this in mind for future development. Feel free to add a `TODO` comment about it 😄

spec/sql_overrides.rb
solnic: For now let's do this:

``` ruby
# FIXME: refactor so that it returns an intermediate function representation which
#              needs to a name
::ROM::SQL::Function.new(type, name: :__anonymous__, schema: schema)
```

I'll make sure this gets addressed before the release.

solnic:
> Besides any change required to the API or the implementation, it would just remain to be done moving other meta's to options, if we think it is a good idea.

Let's move the rest in a separate PR. This is getting big enough already.

spec/sql_overrides.rb
waiting-for-dev: @solnic I was now re-adapting `rom-sql` to new changes, and I was adding `name: :__anonymous` everywhere a function is created without name. The problem is that `AS "__anonymous__"` gets added to a lot of resulting SQL queries.

Maybe we should tackle that refactor first... or just stick with `nil` names in the meanwhile.

solnic: Oh damn :/ My bad, apologies, I didn't think this through. Could you try redefining `option :name, optional: true, type: Types::Symbol` and see if that works?

solnic:
@waiting-for-dev thanks again for working on this. I rebased it and cleaned up commits. I'll merge once CI is green.

waiting-for-dev:
Hey @solnic I was going to tackle CI offenses rigjt now, but it seems you just merged it at the same exact time?

solnic:
@waiting-for-dev feel free to push such fixes directly to master

waiting-for-dev:
Ok @solnic , thanks for your feedback and your support with this PR.

PR #521, opened 2019-04-17
Fix return values mentioned in documentation

lauratpa:
I was going through the documentation and found these return values mentioned, that don't seem to be inline with the code.


solnic:
Good catch. Thank you!

PR #543, opened 2019-05-05
Make Schema#[] raise when attribute name is not unique

### LINKED ISSUE #529: Schema#[] should raise if more than one attribute was found
If a schema contains more than one attribute with the same name, which can happen when you join schemas in sql, using `Schema#[]` will return the first matching attribute. This should be changed to raise an error when more than one attribute was found.

This is a bug fix but in a way it's also a breaking change, so this goes to `6.0.0`.

Refs #524

waiting-for-dev:
Fixes #529

### DIFF
--- a/lib/rom/schema.rb
+++ b/lib/rom/schema.rb
@@ -221,17 +221,16 @@ module ROM
     # @raise KeyError
     #
     # @api public
-    def [](key, src = name.to_sym)
-      attr =
-        if count_index[key].equal?(1)
-          name_index[key]
-        else
-          source_index[src][key]
-        end
-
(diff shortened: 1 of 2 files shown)


waiting-for-dev:
It is all a matter of taste, but if you prefer I can make hounds happy :smile:

solnic:
@waiting-for-dev 👋  any chance you could rebase it on top of latest master? I merged all 4 gems into one so now everything is in `lib` + we made lots of rubocop adjustments.

waiting-for-dev:
Done @solnic :rocket:

PR #545, opened 2019-06-13
Generalize plugin API

flash-gordon:
We should have done it earlier, it's way simpler now.


core/lib/rom/plugin_registry.rb
```diff
@@ -61,48 +27,48 @@ def initialize
...
-      when :command       then commands.adapter(adapter)
-      when :mapper        then mappers.adapter(adapter)
-      when :relation      then relations.adapter(adapter)
-      when :schema        then schemas.adapter(adapter)
+      when :relations then :relation
+      when :commands then :command
+      when :mappers then :mapper
+      when :schemas then :schema
+      else type
       end
```
solnic: @flash-gordon should we maybe use the inflector here?

core/lib/rom/plugin_base.rb
```diff
@@ -19,10 +24,11 @@ class PluginBase
...
-    def initialize(mod, options)
-      @mod      = mod
-      @options  = options
-      @type = options.fetch(:type)
+    def initialize(name, mod, options)
+      @name    = name
+      @mod     = mod
+      @options = options
+      @type    = options.fetch(:type)
     end
```
solnic: Seems like this could just use `Initializer`

core/lib/rom/plugins.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require 'dry/container'
+require 'rom/plugin'
+
+module ROM
+  # Registry of all known plugin types (command, relation, mapper, etc)
+  #
+  # @api private
+  module Plugins
+    @registry = ::Dry::Container.new
```
solnic: why not just `extend Dry::Container::Mixin` and call it a day?

core/lib/rom/plugin_registry.rb
flash-gordon: I thought about it but I see no point, better remove these aliases in the next version.

core/lib/rom/plugins.rb
flash-gordon: no objection against it, I just started small to have a better understanding of what I'll need

flash-gordon:
Oh, actually it's not. Even more, I added name to `Plugin`, this means this method makes no sense now. I'll remove it now.

solnic:
@flash-gordon I recommend paying attention to coverage diffs in PRs. I've had situations like that few times and it helps in spotting dead code.

PR #566, opened 2019-10-07
Fix warning about shadowed arg1

barbeque:
I got this warning when installing `rom-core-5.1.2`:

```
Installing darkfish documentation for rom-core-5.1.2
(eval):9: warning: key :arg1 is duplicated and overwritten on line 9
(eval):9: warning: key :arg1 is duplicated and overwritten on line 9
```

I believe it is related to this line.


flash-gordon:
thanks!

barbeque:
That was super fast! Thanks!

PR #572, opened 2019-12-29
[changeset] update Pipe to use the new `define! do ... end` API

solnic:


solnic:
@flash-gordon this could go into `release-5.2` actually, wdyt?

flash-gordon:
sure, why not, it's still to be released (my schedule is to do it within a week or so)

PR #590, opened 2020-03-01
Use explicit coercion in update changeset

flash-gordon:
I think this way it won't break any existing code around.

### DIFF
--- a/changeset/lib/rom/changeset/update.rb
+++ b/changeset/lib/rom/changeset/update.rb
@@ -65,7 +65,7 @@ module ROM
       def diff
         @diff ||=
           begin
-            source = Hash(original)
+            source = original.to_h
             data = pipe.for_diff(__data__)
             data_tuple = data.to_a
             data_keys = data.keys & source.keys


changeset/lib/rom/changeset/update.rb
```diff
@@ -65,7 +65,7 @@ def clean?
       def diff
         @diff ||=
           begin
-            source = Hash(original)
+            source = original.to_h
```
solnic: @flash-gordon `Hash` expects `to_h` so this is fine BUT now you'd get `NoMethodError` rather than `TypeError` so we should probably maintain this behavior

flash-gordon: @solnic but what's the point? `.to_h` is a less strict constraint, you almost certainly won't get an error during upgrade. Defensive programming doesn't help unless you expect something to fail. And, finally, differentiating between `NoMethodError` and `TypeError` is a joke in a language like ruby.

solnic: OK fair enough

flash-gordon: @solnic to be completely fair, I'm in favor of adding semantics whenever possible. This for sure does include robust using exception types. But when we don't expect a failure it's better not to bother

PR #591, opened 2020-03-01
Allow to configure custom inflector

flash-gordon:
All components that rely on inflector use an injected inflector now rather than using a global `ROM::Inflector`. This makes it possible to configure a custom inflector like that:

```ruby
configuration = ROM::Configuration.new(:memory, "something") do |conf|
  conf.inflector = my_custom_inflector
end
```

### DIFF
--- a/lib/rom/associations/definitions/abstract.rb
+++ b/lib/rom/associations/definitions/abstract.rb
@@ -7,6 +7,7 @@ require "rom/types"
 require "rom/initializer"
 require "rom/relation/name"
 require "rom/associations/through_identifier"
+require "rom/support/inflector"
 
 module ROM
   module Associations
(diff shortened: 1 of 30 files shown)


solnic:
@flash-gordon wouldn't it be better to just have `ROM.config.inflector` with a helper module for all the classes that needs it? It seems overkillish to inject it through options everywhere.

flash-gordon:
@solnic I don't know yet. It's kinda experiment, things like this allow to learn more about your code. Overall, I think inflector _should be_ injectable. This may be not an issue for me personally but for Hanami user/developer relying on a global state is a well-known PITA. Code-wise it seems to be not that bad.

solnic:
@flash-gordon yeah I hear you. It was just an immediate thought I had when I looked at the diff.

flash-gordon:
@solnic apart from failing specs on 2.4 it's ready to go and I believe it's worth it. The changes are not radical and it brings some flexibility. Of course, it won't work when you instantiate two containers with different inflectors but the same relation classes but this case doesn't exist in the real world I'm sure.

solnic:
@flash-gordon I'm sorry but I forgot about this PR and now we have a bunch of conflicts. Do you want this in 5.x or just 6.0?

flash-gordon:
@solnic np, I'll rebase today, it's for 6.0

flash-gordon:
done

PR #592, opened 2020-03-21
[changeset] fix binding map functions in Changeset::Update

timriley:
`map` blocks defined in `ROM::Changeset::Update` subclasses should be evaluated in the context of that changeset's _instance_, however, since the changes in https://github.com/rom-rb/rom/commit/d742f608958a5a6391f53165736f325a64d0066b, they're evaluated in the context of the changeset _class_.

This breaks ROM's documented behaviour and has caused some issues in an application I help with, which depends on the previous (non-broken) behaviour.

Of the 2 tests added here, only the test using an `Update` subclass fails, indicating that this behaviour is somehow specific to changesets using that superclass only.

### DIFF
--- a/changeset/spec/integration/changeset_spec.rb
+++ b/changeset/spec/integration/changeset_spec.rb
@@ -116,6 +116,32 @@ RSpec.describe 'Using changesets' do
       expect(result.updated_at).to be_instance_of(Time)
     end
 
+    it 'preprocesses data using map blocks in a custom class' do
+      changeset_class = Class.new(ROM::Changeset::Create) do
+        map do |tuple|
+          extend_tuple(tuple)
+        end
+
+        private
+
+        def extend_tuple(tuple)
+          tuple.merge(title: tuple[:title] + ", yes really")
+        end
+      end
+
+      changeset = books.changeset(changeset_class, title: 'rom-rb is awesome')
+      command = books.command(:create)
+      result = command.(changeset)
+
+      expect(result.id).to_not be(nil)
+      expect(result.title).to eql('rom-rb is awesome, yes really')
+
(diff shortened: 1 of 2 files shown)


changeset/spec/integration/changeset_spec.rb
```diff
@@ -116,6 +116,32 @@ class User < Dry::Struct
...
+      result = command.(changeset)
+
+      expect(result.id).to_not be(nil)
+      expect(result.title).to eql('rom-rb is awesome, yes really')
+
+      result = books.changeset(changeset_class, title: 'rom-rb is awesome').commit
+
+      expect(result.id).to_not be(nil)
+      expect(result.title).to eql('rom-rb is awesome, yes really')
+    end
```
timriley: This test passes!

changeset/spec/integration/changeset_spec.rb
```diff
@@ -160,6 +186,30 @@ class User < Dry::Struct
...
+      end
+
+      result = books.by_pk(book.id)
+        .changeset(changeset_class, title: 'rom-rb is awesome for real')
+        .commit
+
+      expect(result.id).to be(book.id)
+      expect(result.title).to eql('rom-rb is awesome for real, yes really')
+      expect(result.extended).to be_true
+    end
```
timriley: This test fails, despite being almost identical to the test above (the only difference being it's an `Update` subclass, instead of `Create`).

If you put a `byebug` inside that map block, you'll see output like this:

```
(byebug) self
#<Class:0x00007fa9e6a39840>
```

Whereas if you put the byebug inside map block defined in the `Create` subclass in the test above, it shows:

```
(byebug) self
#<#<Class:0x00007ff5629fb8c0> relation=ROM::Relation::Name(books) data={:title=>"rom-rb is awesome"}>
```

Issue #594, opened 2020-04-01
When writing, custom constructors on schema attribute types are called twice, with values of different types


timriley:
To reproduce this:

1. `createdb rom_double_type_constructor_bug`
2. Save the reproduction script (at the end of this issue) inside the root of the rom checkout
3. Run it: `bundle exec double_type_constructor_bug.rb`

You'll see that the `puts` inside the custom write type constructor is called twice, and the type of the value is different each time, e.g.

```
tags constructor called with Array
tags constructor called with Sequel::Postgres::JSONBArray
```

The _expcted_ behaviour is that this constructor should be called _once only_, and with the value being the regular Ruby class (`Array`), not the coerced-for-sequel class (i.e. Sequel::Postgres::JSONBArray).

The reason the current behavior is problematic is because application authors may (quite reasonably!) be using strict types that crash when encountering the `Sequel` value, e.g.

```ruby
attribute :some_array, ROM::SQL::Types::PG::JSONB.prepend(&Types::Strict::Array.of(MyApp::Types::SomeHashSchema))
```

Which currently crashes like so:

```
*** Dry::Types::SchemaError Exception: [] (Sequel::Postgres::JSONBArray) has invalid type for :some_array violates constraints (type?(Array, []) failed)
```

This behavior is a regression. In 5.0, everything was working fine.

The post-5.0 commit that introduced the erroneous behavior was https://github.com/rom-rb/rom/commit/a80e3fcb8ae5daeeee380b276a404406b79e0227. Specifically, this line: https://github.com/rom-rb/rom/commit/a80e3fcb8ae5daeeee380b276a404406b79e0227#diff-603c977ed0a2443dc61c7d4efbc35b69R196

However, this is what @flash-gordon supposedly fixed in https://github.com/rom-rb/rom/pull/589.

So there must be some aspect of the incorrect behavior still hanging around.

Repro script follows:

```ruby
require "bundler/setup"
require "rom"
require "rom/sql"

# Configure database URL
config = ROM::Configuration.new(:sql, "postgres://localhost/rom_double_attribute_write_type_call")
rom = ROM.container(config)

# Migrate database
rom.gateways[:default].tap do |gateway|
  migration = gateway.migration do
    change do
      drop_table? :articles

      create_table :articles do
        primary_key :id
        column :title, :text, null: false
        column :tags, :jsonb, null: false, default: "[]"
      end
    end
  end

  migration.apply gateway.connection, :up

  # Populate data
[...]

PR #595, opened 2020-04-04
Fix duplicated input processing in update commands

flash-gordon:
It should be backported ofc

### DIFF
--- a/core/lib/rom/plugins/command/schema.rb
+++ b/core/lib/rom/plugins/command/schema.rb
@@ -22,18 +22,18 @@ module ROM
           #
           # @api public
           def build(relation, **options)
-            if relation.schema?
-              input = options.fetch(:input, self.input)
+            if relation.schema? && !options.key?(:input)
               relation_input = relation.input_schema
+              command_input = input
 
-              input_handler =
-                if input.equal?(ROM::Command.input)
+              composed_input =
+                if command_input.equal?(ROM::Command.input)
                   relation_input
                 else
-                  -> tuple { relation_input[input[tuple]] }
+                  -> tuple { relation_input[command_input[tuple]] }
                 end
 
-              super(relation, **options, input: input_handler)
+              super(relation, **options, input: composed_input)
             else
               super
             end
--- a/core/spec/integration/commands/update_spec.rb
+++ b/core/spec/integration/commands/update_spec.rb
@@ -11,7 +11,10 @@ RSpec.describe 'Commands / Update' do
 
   before do
     configuration.relation(:users) do
-      schema(:users) {}
+      schema(:users) do
+        attribute :name, ROM::Types::String
+        attribute :email, ROM::Types::String
+      end
 
       def all(criteria)
         restrict(criteria)
@@ -30,8 +33,7 @@ RSpec.describe 'Commands / Update' do
   it 'update tuples' do
     result = users.update.all(name: 'Jane').call(email: 'jane.doe@test.com')
 
-    expect(result)
-      .to match_array([{ name: 'Jane', email: 'jane.doe@test.com' }])
+    expect(result).to eql([{ name: 'Jane', email: 'jane.doe@test.com' }])
   end
 
   describe '"result" option' do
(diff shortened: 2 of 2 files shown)


timriley:
@flash-gordon Hmm, I've now tried this branch on the application where I found the issue and my write type constructors are still getting called twice, firstly with `Array` and then again with `Sequel::Postgres::JSONBArray`.

To confirm it's not just due to a quirk of the larger application, I've dropped my reproduction script from https://github.com/rom-rb/rom/issues/594 into the app and run it against the app's gem bundle, and it indeed still shows the erroneous behavior:

```
$ bundle exec ruby rom_double_write_bug.rb
tags constructor called with Array
tags constructor called with Sequel::Postgres::JSONBArray
```

Did you test against the error reproduction script when making your adjustments here?

timriley:
Here's a copy of the repro script with bundler/inline enabled to show I'm using this git branch etc.

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }

  gem "pg"
  gem "rake"
  gem "rom", github: "rom-rb/rom", branch: "594-fix-duplicated-inputs-in-restrictible-commands"
  gem "rom-sql"
end

require "rom"
require "rom/sql"

# Configure database URL
config = ROM::Configuration.new(:sql, "postgres://localhost/rom_double_attribute_write_type_call")
rom = ROM.container(config)

# Migrate database
rom.gateways[:default].tap do |gateway|
  migration = gateway.migration do
    change do
      drop_table? :articles

      create_table :articles do
        primary_key :id
        column :title, :text, null: false
        column :tags, :jsonb, null: false, default: "[]"
      end
    end
  end

  migration.apply gateway.connection, :up

  # Populate data
  gateway.connection.tap do |connection|
    connection.execute("INSERT INTO articles (title) VALUES ('Test article')")
  end
end

# Relation
module Relations
  class Articles < ROM::Relation[:sql]
    schema :articles, infer: true do
      attribute :tags, ROM::SQL::Types::PG::JSONB.prepend { |value|
        ::Kernel.puts "tags constructor called with #{value.class}"
        value
      }
    end
  end
end

# Register relation, finalize rom container
config.register_relation Relations::Articles
rom = ROM.container(config)

# Repository with update method
class ArticleRepo < ROM::Repository[:articles]
  def update_tags(tags)
    articles.by_pk(1).changeset(:update, tags: tags).commit
  end
end

# Run the update and witness the constructor being called twice
repo = ArticleRepo.new(rom)
repo.update_tags(%w[new tags here])

# e.g.
# tags constructor called with Array
# tags constructor called with Sequel::Postgres::JSONBArray
```

flash-gordon:
@timriley apologies to the delay, I thought I messed with something and it'll require a longer investigation... But it seems that inlined gemfile doesn't work in the rom environment where a few gems reside in a single repo.
I checked it by adding `gem 'pry'` and looking at the sources:
```ruby
puts ROM::Plugins::Command::Schema::ClassInterface.instance_method(:build).source
```
```ruby
          def build(relation, **options)
            if relation.schema?
              input = options.fetch(:input, self.input)
              relation_input = relation.input_schema

              input_handler =
                if input.equal?(ROM::Command.input)
                  relation_input
                else
                  -> tuple { relation_input[input[tuple]] }
                end

              super(relation, **options, input: input_handler)
            else
              super
            end
          end
```
👆 it's the old code.
When I tested it locally using `./bin/console` setup it worked. What's your Gemfile setup? IIRC it should be like 
```ruby
gem 'rom', github: 'rom-rb/rom', branch: 'branch-name' do
  gem 'rom-core'
  gem 'rom-changeset'
  gem 'rom-repository'
end
```

timriley:
Thanks for this @flash-gordon, I really appreciate you following up again!

So I can confirm this works now!! 🎉 

What I had to do, though, is set up my Gemfile like this:

```
  gem "rom", github: "rom-rb/rom", branch: "594-fix-duplicated-inputs-in-restrictible-commands"
  gem "rom-core", github: "rom-rb/rom", branch: "594-fix-duplicated-inputs-in-restrictible-commands"
  gem "rom-changeset", github: "rom-rb/rom", branch: "594-fix-duplicated-inputs-in-restrictible-commands"
  gem "rom-repository", github: "rom-rb/rom", branch: "594-fix-duplicated-inputs-in-restrictible-commands"
```

For some reason the block form didn't work, at least not with my inlined gemfile script.

PR #603, opened 2020-06-25
Fix database name in docker composer config

waiting-for-dev:
Just that :arrow_up: :smiley:


solnic:
I actually changed it to `rom` on purpose since gems have been merged into one and there's just one spec suite that needs `rom` database (bot repository and changeset specs need it). I actually just noticed that CI still uses `rom_repository` db name so I'm gonna change that too. Sorry for the confusion!

PR #625, opened 2021-02-26
Replace references to transproc by dry-transformer

smaximov:
`transproc` was replaced by `dry-transformer`, but there were still some references to it left in the docs.


solnic (approved):
Thanks!

PR #628, opened 2021-04-07
Adds hint for auto-registration with nested namespaces

mgiagante:
Auto-registration already supports nesting namespaces, and those are accepted by ROM::AutoRegistration if specified properly with "::" in the namespace string.
This proposal is for documenting this use case.


solnic:
Thanks!

PR #629, opened 2021-04-21
Improved usage example for custom mappers.

mgiagante:
This pull request replaces the current usage example for custom mappers by one with a transforming pipeline and comments on what the functions in the pipeline do.

### DIFF
--- a/core/docsite/source/mappers.html.md
+++ b/core/docsite/source/mappers.html.md
@@ -54,21 +54,16 @@ class UserMapper < ROM::Transformer
     create_instance
     # Any other functions you need.
   end
-
+  
   # Find the model class for the row based on the content of its role
   # field, then add it to the row's data for next function to use it.
   def resolve_model(row)
-    model_for = ->(row) do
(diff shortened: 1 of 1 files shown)


core/docsite/source/mappers.html.md
```diff
@@ -45,14 +45,36 @@ Typically though, custom mappers will be used in more complex cases, when the un
...
+  # field, then add it to the row's data for next function to use it.
+  def resolve_model(row)
+    model_for = ->(row) do
+      Object.const_get(Dry::Inflector.new.camelize(row[:role]))
+    end
 
-  map_array do
-    # define custom transformations here
+    row.to_h.merge(model: model_for.(row))
+  end
```
solnic, line 66: I just realized this could be simplified like this:

```ruby
def resolve_model(row)
  [Inflector.classify(row[:role]), row]
end

def create_instance(model, row)
  model.new(row)
end
```

Notice `Inflector` which is a constant that we have under `ROM::Inflector`

solnic (changes requested):
Thanks for the PR. Please see my comment 🙂

PR #634, opened 2021-06-07
Use Zeitwerk for loading components

solnic:
This replaces AutoRegistration with a Loader that uses Zeitwerk. I'll open a separate PR once this is merged with the actual API improvements, what this PR does essentially, is providing a better and simpler implementation for loading components. Things like options and behavior still need to be improved.

### TODO

- [x] Figure out how to "reset" Zeitwerk in specs because it keeps a global Registry which breaks specs when you run them all
- [x] Refactor `AutoRegistration` into a component loader ~where each instance handles a specific component type~ turned out a single loader is needed because zeitwerk can't handle multiple dirs that have the same parent
- [x] Refactor `Setup#auto_registration` into `Setup#auto_register` and refine its behavior to use the new loader
- [x] Figure out if Zeitwerk can support passing a namespace that is not part of a file path. ie `lib/relations/users.rb` where the constant is actually `MyNamespace::Users` - ~~it cannot and this is fine, because this is a weird use case that's completely non-standard so I'll move this behavior to a backward-compatibility "layer"~~ silly me missed the fact that Zeitwerk actually supports passing "namespace" option to push_dir, so this use case is supported after all
- [x] add `rom/compat` for backward-compatibility APIs. ~Something like `ROM.include(ROM::Compat[5.0])` should suffice~ sticking to plain `require "rom/compat"` for now
- [x] improve `on_load` handling because it actually yields path too, so we can infer component type from this easily
- [x] address RuboCop complaints
- [x] clean up commits


lib/rom/setup.rb
```diff
@@ -31,57 +16,76 @@ class Setup
...
     def register_plugin(plugin)
       plugins << plugin
     end
+
+    private
+
+    # @api private
+    def loader
+      @loader ||= Loader.new(@auto_register[0], **(@auto_register[1] || {})) if @auto_register
+    end
```
solnic, line 89: The way Loader is handled along with this components hash is going to be refactored separately. So, please ignore this part here.

lib/rom/loader.rb
```diff
@@ -0,0 +1,153 @@
...
+    extend Initializer
+
+    NamespaceType = Types::Strict::Bool | Types.Instance(Module)
+
+    PathnameType = Types.Constructor(Pathname, &Kernel.method(:Pathname))
+
+    InflectorType = Types.Interface(:camelize)
+
+    ComponentDirs = Types::Strict::Hash.constructor { |hash|
+      hash.map { |key, value| [key, value.to_s] }.to_h
```
flash-gordon, line 26: `hash.to_h { |key, value| [key, value.to_s] }` or `hash.transform_values(&:to_s)`?

flash-gordon (approved):

PR #645, opened 2021-07-03
Introduce gateway component

solnic:
This replaces immediate loading of gateways during configuration phase
with the component-based approach which makes it lazy-loadable.

Backward-compatibility is kept via rom/compat, if you don't use it your
gateways will be loaded upon first access within the runtime container.

Here's an example:

```ruby
# Doing this no longer connects to the db
rom = ROM.container(:sql, "postgres://localhost/my_db")

# The gateway will be initialized and connected once you access it for the first time.
# This means that whenever a relation loads and asks for its dataset, the corresponding
# gateway will be initialized
rom.gateways[:default]
```


lib/rom/compat.rb
```diff
@@ -99,10 +100,23 @@ def relation_classes(gateway = nil)
...
+    # @api public
+    # @deprecated
+    def [](key)
+      gateways.fetch(key)
+    end
+
+    # @api public
+    # @deprecated
+    def gateways
+      @gateways ||= setup.components.gateways.map(&:build).map { |gw| [gw.config.name, gw] }.to_h
```
flash-gordon, line 112: `to_h` yields block starting 2.6 https://ruby-doc.org/core-2.6/Enumerable.html#method-i-to_h

solnic: @flash-gordon hey sorry I missed this. I will open "a final PR" against "a virtual" branch once I'm done and happy with the outcome, so that you can go through diff and comment on things. Then I can address all-the-feedback. It's gonna be a massive diff of course but the code should be way more understandable than before.

PR #658, opened 2021-09-05
Import Transproc::Conditional

ippachi:
Hi 😃 
I want to use Transproc::Conditional, so I added it


ianks:
Thanks for your contribution!

PR #686, opened 2023-09-28
Fix `command(:create, result: :many)` for combines

dgollahon:
- In ruby 3.x the `result: :many` will get forwarded incorrectly if there is a `combine`. That is, something like `relation.command(:create, result: :many)` will work but `relation.combine(:child_relation).command(:create, result: :many)` will raise `ArgumentError: wrong number of arguments (given 2, expected 1)`. This fixes is it (I have tested by doing the following on my codebase `ROM::Relation::Combined.__send__(:ruby2_keywords, :command)`).

### DIFF
--- a/lib/rom/relation/combined.rb
+++ b/lib/rom/relation/combined.rb
@@ -148,6 +148,7 @@ module ROM
                 "#{self.class}#command doesn't work with #{type.inspect} command type yet"
         end
       end
+      ruby2_keywords(:command) if respond_to?(:ruby2_keywords, true)
 
       private
 


dgollahon:
NOTE: I have not added any tests here because it was too much effort and I had too little time to figure out how to set up a local environment that worked. I would appreciate either someone else adding a test or pointers for getting set up.

flash-gordon:
I think we can drop ruby 2.x safely

katafrakt:
I see 2.x was already dropped from rom-sql. So should we do the same here? I can create the PR if needed.

dgollahon:
> I see 2.x was already dropped from rom-sql. So should we do the same here? I can create the PR if needed.

To be clear, this isn't directly an issue of 2.x support. This doesn't work on 3.x, using 2.x style keywords happens to fix it.

solnic:
Thanks for the PR! So you're saying that without this, it's broken under 3.x?

solnic:
> I think we can drop ruby 2.x safely

Yes let's do this

dgollahon:
> Thanks for the PR! So you're saying that without this, it's broken under 3.x?

Correct, when combined with `#combine`.

solnic:
How about we just ditch 2.x and fix signatures to be `command(type, **opts)`?

Issue #610, opened 2020-09-15
Being able to initialise a fully populated ROM::Struct


mereghost:
`ROM::Struct`s have albeit having a schema (defined by query returns if I got it right) don't allow you to use them for initialisation.

```
# table authors (name:string, id:integer, created_at:datetime, updated_at:datetime)
class Author < ROM::Struct ; end
Author.new(name: 'Bob') # => <Author {}>
```

This makes it hard to test pieces of your code without touching the DB at all and therefore slowing the testing down. Is it possible to offer some similar functionality (maybe as part of `rom-sql` :/ )?

```
Author.new(name: 'Bob') # => <Author {name: 'Bob'}>
```

flash-gordon:
@mereghost rom-factory does that, have you seen it?

solnic:
@flash-gordon the idea is to be able to do it via class constant. Thoughts?

flash-gordon:
@solnic type metadata will be missing, it can be a problem in some cases. For instance, you won't be able to notice column renaming. Otherwise, it's possible with `method_missing` and overriding of `#to_s`

flash-gordon:
though I think I would go with a different name, `Struct.build` or something

solnic:
> @solnic type metadata will be missing, it can be a problem in some cases. For instance, you won't be able to notice column renaming

@flash-gordon not sure if I follow - we'd use actual struct classes generated by the struct compiler, they have meta-data. I'd imagine that if you do `User.build(name: "Jane")` then the factory would search for the canonical representation of that struct class (basically going to `users` relation and grabbing its `users.mapper.model`) and using that. So, essentially a little bit of magic behind the scenes.

I like `Struct.build` btw.

flash-gordon:
@solnic linking `User` to a container would require some sort of a global state, this concerns me somewhat :)

solnic:
@flash-gordon but it's already "linked" through `struct_namespace` and class name inference, isn't it?

flash-gordon:
@solnic no, it's the other way around, container references classes. Currently, `User` is an empty class that can be easily shared by multiple containers with (possibly) different schemas

solnic:
@flash-gordon you're right. I guess we should stick to factories after all. @mereghost would you be OK with that? Please remember that `User` is only a placeholder class, the actual struct classes used at runtime are subclasses of that `User` class and they are created on-the-fly, based on relation schemas.

mereghost:
[...]

PR #693, opened 2024-06-12
Fix ROM Struct examples

### LINKED ISSUE #685: Broken example code, unexpected auto_struct behavior
## Describe the bug

I made a script of the example code, and it seems that the example code for the auto-generated structs no longer reflects reality, and as a result I am very confused about how to setup my custom name-spaced structs, when not even the auto-generated ones work as expected.

The example code I mostly copied is here: https://api.rom-rb.org/rom/ROM/Struct
Note that as written the example code doesn't work _at all_, as it was missing the `conf.relation(:users)`, so I added that.  Even with that, the example code still doesn't work at all, and raises the error:
`undefined method 'schema' for ROM::OpenStruct:Class (NoMethodError)`

## To Reproduce

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rom'
  gem 'rom-sql'
  gem 'sqlite3'
end

rom = ROM.container(:sql, 'sqlite::memory') do |conf|
  conf.default.create_table(:users) do
    primary_key :id
    column :name, String
  end

  conf.relation(:users) do
  end
end

class UserRepo < ROM::Repository[:users]
end

user_repo = UserRepo.new(rom)

# get auto-generated User struct
model = user_repo.users.mapper.model

puts "model: #{model}"
# EXPECTED
# model: ROM::Struct::User
#
# ACTUAL
# model: ROM::OpenStruct

puts model.schema.key(:id)
# EXPECTED:
# #<Dry::Types[id: Nominal<Integer meta={primary_key: true, source: :users}>]>
#
# ACTUAL:
# undefined method `schema' for ROM::OpenStruct:Class (NoMethodError)
```

## My environment

- Affects my production application: **NOT YET** (Implementing ROM as part of a new service)
- Ruby version: 2.7.8 through 3.2.2
- OS: macOS Ventura 13.4

cllns:
Fixes #685

Also adds a **NOTE** in front of the caveat for ROM structs which was also discussed in #610

### DIFF
--- a/core/lib/rom/struct.rb
+++ b/core/lib/rom/struct.rb
@@ -17,9 +17,10 @@ module ROM
   # about attribute types returned from relations, thus can be introspected to build
   # additional functionality when desired.
   #
-  # There is a caveat you should know about when working with structs. Struct classes
-  # have names but at the same time they're anonymous, i.e. you can't get the User struct class
-  # with ROM::Struct::User. ROM will create as many struct classes for User as needed,
+  # *NOTE*: There is a caveat you should know about when working with ROM structs.
+  # Struct classes have names but at the same time they're anonymous,
+  # i.e. you can't get the User struct class with ROM::Struct::User.
+  # ROM will create as many struct classes for User as needed,
   # they all will have the same name and ROM::Struct::User will be the common parent class for
   # them. Combined with the ability to provide your own namespace for structs this enables to
   # pre-define the parent class.
--- a/core/lib/rom/struct.rb
+++ b/core/lib/rom/struct.rb
@@ -31,11 +31,14 @@ module ROM
   #       primary_key :id
   #       column :name, String
   #     end
+  #
+  #     conf.relation(:users) do
+  #       schema(infer: true)
+  #       auto_struct true
+  #     end
   #   end
   #
   #   class UserRepo < ROM::Repository[:users]
-  #     schema(infer: true)
-  #     auto_struct true
   #   end
   #
   #   user_repo = UserRepo.new(rom)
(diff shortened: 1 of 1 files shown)


core/lib/rom/struct.rb
```diff
@@ -33,6 +34,8 @@ module ROM
   #   end
   #
   #   class UserRepo < ROM::Repository[:users]
+  #     schema(infer: true)
+  #     auto_struct true
```
flash-gordon, line 38: these are settings for relations, not repos

core/lib/rom/struct.rb
```diff
@@ -43,11 +46,11 @@ module ROM
   #
   #   # see struct's schema attributes
   #
-  #   # model.schema.key(:id)
-  #   # => #<Dry::Types[id: Nominal<Integer meta={primary_key: true, source: :users}>]>
+  #   model.schema.key(:id)
+  #   # => #<Dry::Types[id: Nominal<Integer meta={primary_key: true, alias: nil, source: :users}>]>
```
flash-gordon, line 50: I think omitting insignificant meta keys is fine here

core/lib/rom/struct.rb
cllns: Ah, apologies. I made sure to test the whole script but put it in the wrong place when adding it to the comment. Fixed now

core/lib/rom/struct.rb
cllns: Gotcha, just removed `alias`

flash-gordon (approved):

flash-gordon:
thank you!

Ahmadtaha1122:
👍

PR #695, opened 2024-07-20
Reduce the retained memory when defining the relation readers

### LINKED ISSUE #694: Excessive memory use with many relations (>500)
## Describe the bug

When booting with a lot of relations (we have 580) and profiling the setup of the container using [vernier](https://github.com/jhawthorn/vernier) we see a lot of retained memory around the `registry_reader` plugin

https://github.com/rom-rb/rom/blob/v5.3.2/core/lib/rom/setup/finalize/finalize_relations.rb#L46

It looks as though the plugin used to setup the relation methods on each of the relations is causing the issue

### Call Tree

image

### Flame graph

image

After experimenting I can reduce this quite considerably over using one of two techniques (366Mb -> 155Mb)

1. Define the methods in a anonymous module and then include that into each relations by passing it into the plugin
https://gist.github.com/DangerDawson/0eb82f7e61cc791a31af120427f128ac

2. The same as 1. but use a class attribute to cache the anonymous module
https://gist.github.com/DangerDawson/46164ba54de58e8a697d2ecb9f36a02e

### Call Tree

image

### Flame graph

image

I am happy to submit a patch, but wanted to get an idea of the preference of the two solutions.

I have also seen the same issue with the repositories as well, which take a similar approach to setup the relations for each repository, but I will follow up with that once I have had some feedback on the above

As a footnote, even with the patch does over 150mb or retained memory sound reasonable? I keep thinking we may have some references to some large objects which we are not freeing up after the relations are built

DangerDawson:
Fixes: https://github.com/rom-rb/rom/issues/694

### DIFF
--- a/core/lib/rom/plugins/relation/registry_reader.rb
+++ b/core/lib/rom/plugins/relation/registry_reader.rb
@@ -14,12 +14,8 @@ module ROM
         EMPTY_REGISTRY = RelationRegistry.build(EMPTY_HASH).freeze
 
         # @api private
-        attr_reader :relations
-
-        # @api private
-        def initialize(relations:)
-          @relations = relations
-          define_readers!
+        def initialize(klass:, relation_readers_module:)
+          klass.include relation_readers_module
         end
 
         # @api private
@@ -28,15 +24,6 @@ module ROM
           return if klass.instance_methods.include?(:__registry__)
 
           klass.option :__registry__, default: -> { EMPTY_REGISTRY }
-        end
-
-        private
-
-        # @api private
-        def define_readers!
-          relations.each do |name|
-            define_method(name) { __registry__[name] }
-          end
         end
       end
     end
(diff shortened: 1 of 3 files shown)


solnic (approved):
LGTM thanks for working on this!

solnic:
@DangerDawson seems like repository specs are now failing :/ looks like "Optimise the repository relations reader" is the culprit.

DangerDawson:
It looks as though it is something to do with the way I am caching the `build_relation_readers` in the class attribute.

I will try and carve sometime out over the weekend to understand what is going on, bit it feels like state leaking between the specs as the specs that fail pass when they are ran individually

solnic:
Yeah that's what I suspected. We can have an after callback that clears the cache in RSpec.

DangerDawson:
This is probably not the most elegant way, maybe using dry configurable and having a callback reset the class attributes back to their defaults would be nicer. But I wanted provide a fix ASAP in order to get the branch green again:

https://github.com/rom-rb/rom/pull/696

